PR #38, opened 2021-06-24
fix: fixes #25 correctly set of dhcp + nil check +  firewall_type in server

### LINKED ISSUE #25: Wrong setting for cloudapiv6 k8s-lans-dhcp after terraform apply.
## Description
We use the new feature of static routes for our Kubernetes clusters. This feature was added in the IONOS Cloud API v6. The problem now is that we urgently want to disable "dhcp". Do I set this directly in the Cloud API v6 via an HTTPS request. Is that also taken over correctly. However, the Terraform provider does not do this. 

## Expected behavior
The expected behavior would be that if I set dhcp = false in the "lans" property in the object in the K8s node pool, but the IONOS cloud maintains this value at "true". As a result, our "internal" network is not reachable. 

## Environment

Terraform version: 
```
0.14.3
```
Provider version: 
```
6.0.0-alpha.2
```
OS:
```
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
```

## How to Reproduce
This can be easily reproduced by using the new Terraform Provider with v6 support and setting the resource `ionoscloud_k8s_node_pool` to `false` for the LANs DHCP. After a Terraform apply this value will not be applied to the IONOS cloud.
```
resource "ionoscloud_k8s_node_pool" "np0" {
  availability_zone = "ZONE_1"
  cores_count       = 8
  cpu_family        = "INTEL_SKYLAKE"
  datacenter_id     = ionoscloud_datacenter.dc.id
  k8s_cluster_id    = ionoscloud_k8s_cluster.ncc.id
  k8s_version       = "1.20.6"
  lans {
    id   = ionoscloud_lan.internal.id
    dhcp = false
```

## Error and Debug Output
Statefile:
```
            "k8s_version": "1.20.6",
            "lans": [
              {
                "dhcp": true,
                "id": 1
```

## Additional Notes
That issue on the ionos-cloud/sdk-go looks a lot like the cause to me:
https://github.com/ionos-cloud/sdk-go/issues/5

### LINKED ISSUE #40: IpFailover is missing in lan from create and update functions
## Description
Although ip_failover field is present in schema as optional, it has been implemented only in the read function, missing from the create and update functions

## Expected behavior
ip_failover should be only computed in lan 

## Environment

Terraform version: 
```
0.15.3
```
Provider version: 
```
v5.2.2, v6.0.0-alpha.3
```
OS:
```
OSX 11.2.3, Linux, Windows
```

IuliaBlindu:
- fixes #25 correctly set of dhcp - added true as default and replaced terraform getOk function with get
- nil check for nodePoolLan.Routes 
- added firewall_type field in nic entity of server resource
- docs update
- fix #40 ip_failover just computed for lan

### DIFF
--- a/docs/resources/k8s_node_pool.md
+++ b/docs/resources/k8s_node_pool.md
@@ -20,7 +20,10 @@ resource "ionoscloud_k8s_node_pool" "demo" {
     min_node_count = 1
     max_node_count = 3
   }
-  lans = [1, 2, 3]
+  lans {
+    id   = ionoscloud_lan.terraform_acctest.id
+    dhcp = false
+   }
   maintenance_window {
     day_of_the_week = "Sunday"
     time            = "10:30:00Z"
@@ -46,7 +49,7 @@ The following arguments are supported:
 - `name` - (Required)[string] The name of the Kubernetes Cluster.
 - `k8s_version` - (Optional)[string] The desired Kubernetes Version. for supported values, please check the API documentation.
 - `auto_scaling` - (Optional)[string] Wether the Node Pool should autoscale. For more details, please check the API documentation
-- `lans` - (Optional)[list] A list of numeric LAN id's you want this node pool to be part of. For more details, please check the API documentation, as well as the example above
+- `lans` - (Optional)[list] A list of additional LANs attached to worker nodes. For more details, please check the API documentation, as well as the example above
 - `maintenance_window` - (Optional) See the **maintenance_window** section in the example above
 - `datacenter_id` - (Required)[string] A Datacenter's UUID
 - `k8s_cluster_id`- (Required)[string] A k8s cluster's UUID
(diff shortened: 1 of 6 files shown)


docs/resources/k8s_node_pool.md
```diff
@@ -46,7 +49,7 @@ The following arguments are supported:
 - `name` - (Required)[string] The name of the Kubernetes Cluster.
 - `k8s_version` - (Optional)[string] The desired Kubernetes Version. for supported values, please check the API documentation.
 - `auto_scaling` - (Optional)[string] Wether the Node Pool should autoscale. For more details, please check the API documentation
-- `lans` - (Optional)[list] A list of numeric LAN id's you want this node pool to be part of. For more details, please check the API documentation, as well as the example above
+- `lans` - (Optional)[list] A list of additional LANs attached to worker nodes. For more details, please check the API documentation, as well as the example above
```
mflorin, line 52: Please also document the default value for `lans[i].dhcp`

ionoscloud/resource_k8s_node_pool.go
```diff
@@ -269,10 +270,10 @@ func resourcek8sNodePoolCreate(ctx context.Context, d *schema.ResourceData, meta
 					lan.Id = &lanID
 					addLan = true
 				}
-				if lanDhcp, lanDhcpOk := d.GetOk(fmt.Sprintf("lans.%d.dhcp", lanIndex)); lanDhcpOk {
-					lanDhcp := lanDhcp.(bool)
-					lan.Dhcp = &lanDhcp
-				}
+
```
mflorin, line 273: Please also update the check in the `Update` function below, where there's still 
```
if lanDhcp, lanDhcpOk := d.GetOk(fmt.Sprintf("lans.%d.dhcp", lanIndex)); lanDhcpOk {
					lanDhcp := lanDhcp.(bool)
					lan.Dhcp = &lanDhcp
				}
```
which will fail when `Dhcp` is false.

ionoscloud/resource_k8s_node_pool_test.go
```diff
@@ -202,3 +224,43 @@ resource "ionoscloud_k8s_node_pool" "terraform_acctest" {
...
+}
+
+resource "ionoscloud_k8s_node_pool" "terraform_acctest" {
+  name              = "%s"
+  datacenter_id     = ionoscloud_datacenter.terraform_acctest.id
+  k8s_cluster_id    = ionoscloud_k8s_cluster.terraform_acctest.id
+  k8s_version       = ionoscloud_k8s_cluster.terraform_acctest.k8s_version
+  lans {
+    id   = ionoscloud_lan.terraform_acctest.id
+    dhcp = false
```
mflorin, line 257: Please include a test for checking the update from `dhcp = true` to `dhcp = false` to make sure everything is fine in the `Update` function with regards to `fetching` the correct value for the dhcp field.

mflorin (changes requested):

PR #49, opened 2021-07-02
Fixed tests + docs update + merged fixes

IuliaBlindu:
- merged fix for correctly setting the active property of the s3 key upon creation
- fix of NetworkLoadBalancerForwardingRule resource at setting health_check
- updated documentation and fixed old one
- merged nic fix to accept list of ips
- fixed group and server tests
- updated image in tests to be more generic not to have to update the tests so frequently 
- replaced apiResponse.Status.StatusCode with apiReponse.StatuCode since the previously form was redundant and sometimes apiRsponse.Status was nil

### DIFF
--- a/docs/data-sources/lan.md
+++ b/docs/data-sources/lan.md
@@ -1,6 +1,6 @@
 ---
 layout: "ionoscloud"
-page_title: "ProfitBricks : ionoscloud_lan"
+page_title: "IonosCloud : ionoscloud_lan"
 sidebar_current: "docs-ionoscloud-datasource-lan"
 description: |-
 Get information on a Ionos Cloud Lans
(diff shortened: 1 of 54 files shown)


docs/data-sources/lan.md
```diff
@@ -14,16 +14,18 @@ The lans data source can be used to search for and return existing lans.
 
 ```hcl
 data "ionoscloud_lan" "lan_example" {
-  name     = "My Lan"
+  datacenter_id = ionoscloud_datacenter.example.id
+  name			= "example_"
```
AnaMariaMoga, line 18: I think it would be great, that at some point, to start using `gofmt` for code formatting. Right now, it would probably be a lot of changes to add and it is not a priority.

AnaMariaMoga (approved):

PR #54, opened 2021-07-16
fixed import for server (+go fmt)

IuliaBlindu:


ionoscloud/utils.go
```diff
@@ -28,25 +28,188 @@ func resourceResourceImport(_ context.Context, d *schema.ResourceData, _ interfa
...
+	if ok && len(bootVolume.(string)) > 0 {
+		_, _, err = client.ServerApi.DatacentersServersVolumesFindById(ctx, datacenterId, d.Id(), bootVolume.(string)).Execute()
+		if err != nil {
+			if err := d.Set("volume", nil); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if server.Properties.BootCdrom != nil {
```
mflorin, line 208: nil test performed on `server.Properties.BootCdrom` but we're dereferencing `server.Properties.BootCdrom.Id`, please add a nil check on `Id` as well.

mflorin (approved):

PR #56, opened 2021-07-16
Feat/dataport security

IuliaBlindu:


ionoscloud/resource_k8s_cluster.go
```diff
@@ -103,6 +103,31 @@ func resourcek8sCluster() *schema.Resource {
...
+			},
+			"s3_buckets": {
+				Type:        schema.TypeList,
+				Description: "List of S3 bucket configured for K8s usage. For now it contains only an S3 bucket used to store K8s API audit logs.",
+				Optional:    true,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
```
mflorin, line 124: what is the reason for not having a list of strings here, are we planning for more properties on the `s3_buckets` items?

PR #62, opened 2021-07-23
fix(server): add user_data to inline volume schema

simt2:
The embedded `volume` in `ionos_server` does not accept `user_data` for cloud-init configuration yet. This simple fix works for me against the staging API. 

I'm not entirely sure what's your approach to sanity checking configuration at this point is, so please feel free to add or suggest checks against `image_name` like  the ones in `resource_volume.go`. Also the v6 branch does not build for me without running `go mod tidy && go mod vendor` but I didn't include this in this PR.

### DIFF
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -193,6 +193,11 @@ func resourceServer() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
+						"user_data": {
+							Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


IuliaBlindu:
Thank you for your contribution - it so happens that we were in the middle of working on this, as part of a bigger change. We will make a release soon with these fixed

IuliaBlindu:
@simt2 please check v6.0.0-beta.2

PR #63, opened 2021-07-23
Fix/build

IuliaBlindu:
- updated dependencies
- added missing arguments in server, nic and volume


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 6.0.0-beta-2
```
mflorin, line 1: we should have a dot instead of hyphen i.e. `beta.2`

mflorin (approved):

PR #91, opened 2021-08-23
Fix/missing data sources

IuliaBlindu:

Implemented data source for backup unit resource

## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [x] Changelog updated and version incremented

### DIFF
--- a/ionoscloud/data_source_backup_unit.go
+++ b/ionoscloud/data_source_backup_unit.go
@@ -1,0 +1,144 @@
+package ionoscloud
+
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	ionoscloud "github.com/ionos-cloud/sdk-go/v5"
+	"strings"
+)
+
+func dataSourceBackupUnit() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceBackupUnitRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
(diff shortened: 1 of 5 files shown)


ionoscloud/data_source_backup_unit.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		if backupUnits.Items != nil {
+			for _, bu := range *backupUnits.Items {
+				tmpBackupUnit, _, err := client.BackupUnitApi.BackupunitsFindById(ctx, *bu.Id).Execute()
+				if err != nil {
+					return diag.FromErr(fmt.Errorf("an error occurred while fetching backup unit with ID %s: %s", *bu.Id, err.Error()))
+				}
+				if tmpBackupUnit.Properties.Name != nil {
+					if strings.Contains(*tmpBackupUnit.Properties.Name, name.(string)) {
```
mflorin, line 79: We should go with an exact name match here.

ionoscloud/data_source_backup_unit.go
```diff
@@ -0,0 +1,144 @@
...
+		/* search by ID */
+		backupUnit, _, err = client.BackupUnitApi.BackupunitsFindById(ctx, id.(string)).Execute()
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("an error occurred while fetching the backup unit %s: %s", id.(string), err))
+		}
+	} else {
+		/* search by name */
+		var backupUnits ionoscloud.BackupUnits
+
+		backupUnits, _, err := client.BackupUnitApi.BackupunitsGet(ctx).Execute()
```
mflorin, line 66: Let's try to use a `filter.name={name}&depth=1` GET query here. Not sure the sdk will accept a `filter.name` param though, since it's not in the swagger spec.

PR #100, opened 2021-09-13
correctly importing private lans from k8s node pools

mflorin:

Fixes k8s node pool import when it comes to reading private lans.

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [x] Changelog updated and version incremented


IuliaBlindu (commented):
I think getK8sNodePoolLans could be used also in data_source_k8s_node_pool

PR #108, opened 2021-10-06
SDK-475 add s3 key data source

cristiGuranIonos:
Adds s3 key data source.
## Checklist
- [X] Tests added or updated
- [X] Documentation updated

### DIFF
--- a/ionoscloud/data_source_s3_key_test.go
+++ b/ionoscloud/data_source_s3_key_test.go
@@ -1,0 +1,70 @@
+package ionoscloud
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"testing"
+	"time"
+)
+
+var email = fmt.Sprintf("terraform_test-%d@mailinator.com", time.Now().Unix())
+
+func TestAccDataSourceS3Key_matchFields_expectSuccess(t *testing.T) {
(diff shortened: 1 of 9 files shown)


ionoscloud/data_source_s3_key_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"testing"
+	"time"
+)
+
+var email = fmt.Sprintf("terraform_test-%d@mailinator.com", time.Now().Unix())
+
+func TestAccDataSourceS3Key_matchFields_expectSuccess(t *testing.T) {
```
IuliaBlindu, line 12: Here I would remove the expectSuccess part, because all our test will have this outcome. In the future we will want to integrate them in a CI pipeline so failure would not be our happy outcome :)) 

If there is a certain reason that I don't know (not having so much experience) please let me now

IuliaBlindu (approved):

ionoscloud/data_source_s3_key_test.go
cristiGuranIonos: Old habit since I was writing tests that expected errors(to test all ways a function could fail). Will remove, thanks!

PR #139, opened 2021-11-05
SDK-552 Add share datasource

cristiGuranIonos:

Add share datasource.

## Checklist

- [X] Tests added or updated
- [X] Documentation updated
- [X] Changelog updated and version incremented

### DIFF
--- a/ionoscloud/provider.go
+++ b/ionoscloud/provider.go
@@ -71,7 +71,7 @@ func Provider() *schema.Provider {
 			ServerResource:             resourceServer(),
 			VolumeResource:             resourceVolume(),
 			GroupResource:              resourceGroup(),
-			"ionoscloud_share":         resourceShare(),
+			shareResource:              resourceShare(),
 			UserResource:               resourceUser(),
 			SnapshotResource:           resourceSnapshot(),
 			"ionoscloud_ipfailover":    resourceLanIPFailover(),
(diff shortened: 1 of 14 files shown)


ionoscloud/resource_nic_test.go
```diff
@@ -164,7 +164,7 @@ resource "ionoscloud_nic" "database_nic" {
   lan = 2
   dhcp = false
   firewall_active = false
-  ips            = [ ionoscloud_ipblock.test_server.ips[0], ionoscloud_ipblock.test_server.ips[1] ]
+  ips = [ ionoscloud_ipblock.test_server.ips[0], ionoscloud_ipblock.test_server.ips[1] ]
```
IuliaBlindu, line 167: Not in this PR purpose, but here i would also tests (if the field is updatable) if you can empty the ips list or add one more ip

ionoscloud/provider.go
```diff
@@ -71,7 +71,7 @@ func Provider() *schema.Provider {
 			ServerResource:             resourceServer(),
 			VolumeResource:             resourceVolume(),
 			GroupResource:              resourceGroup(),
-			"ionoscloud_share":         resourceShare(),
+			shareResource:              resourceShare(),
```
IuliaBlindu, line 74: I think we should be consistent regarding naming - ShareResource instead of shareResource

IuliaBlindu (approved):

ionoscloud/provider.go
cristiGuranIonos: That would mean exporting the constant to other packages. The question is, do we want to do that?

ionoscloud/resource_nic_test.go
cristiGuranIonos: good point, will add tests for these

ionoscloud/provider.go
cristiGuranIonos: Will change to uppercase, maybe we will need these in other packages

Issue #144, opened 2021-11-10
Setting firewall rules using icmp_type=0 or icmp_code=0 is broken / Statefiles are not usable with newest version


Z3po:
## Description
When firewall rules using icmp_type, and icmp_code attributes are being managed through the ionoscloud terraform provider it stopped working in version 5.0.4 because of changes introduced in https://github.com/ionos-cloud/terraform-provider-ionoscloud/commit/9b3c28607adea9851c675e65a83ceb4f38f46c71 as those attributes were partly changed in their type.
Now the transformation to `TypeInt` has been continued in https://github.com/ionos-cloud/terraform-provider-ionoscloud/commit/2672a7031bff254d35928bd25e6e66698ed77b70 but unfortunately no state migration mechanism has been introduced.
How are statefiles supposed to be migrated to work with new versions?

Additionally when introducing the changes, the properties are being queried using d.GetOk instead of d.GetOkExists (see https://github.com/ionos-cloud/terraform-provider-ionoscloud/blob/master/ionoscloud/resource_firewall.go#L352-L361) which is not handling 0 type values. This causes resources being created with icmp_type or icmp_code set to 0 being rendered broken with no values being set.

![icmp_type_broken_5_2_18](https://user-images.githubusercontent.com/1095522/141096598-90048ab2-dbdf-4cee-8bbf-c292bd9090ec.png)

## Expected behavior
Old statefiles created with TypeString are being migrated to TypeInt and icmp firewall rules are successfully applied via the API.

This is what it looks like when resources are being applied with version 5.0.4.
![icmp_type_working_5_0_4](https://user-images.githubusercontent.com/1095522/141096711-bc817832-9e3d-4b5a-b21f-af5389761a12.png)

## Environment

Terraform version: 
```
❯ terraform version
Terraform v0.15.3
on linux_amd64
+ provider registry.terraform.io/ionos-cloud/ionoscloud v5.2.18

Your version of Terraform is out of date! The latest version
is 1.0.10. You can update by downloading from https://www.terraform.io/downloads.html
```
Provider version: 
```
❯ terraform providers

Providers required by configuration:
.
└── provider[registry.terraform.io/ionos-cloud/ionoscloud] 5.2.18

Providers required by state:

    provider[registry.terraform.io/ionos-cloud/ionoscloud]
```
OS:
```
❯ cat /etc/issue
Debian GNU/Linux 11 \n \l
```

## Configuration Files
``` terraform
terraform {
[...]

PR #147, opened 2021-11-18
Fix/revert icmp type and icmp code to string

cristiGuranIonos:

Reverts icmp_code and icmp_type from int back to string. Terraform has a problem with unseting integer variables. Even GetOkExists will just set '0' when removing a field, instead of setting it to null. For these fields, null values are valid so we would rather keep them as string, at least until Terraform adds a reliable way to unset int fields.

Note: No upgrade state migration function provided, as the changes were not in production very long.

Note2: There is still a bug in Go sdk that prevents unsetting a field. This is being looked at.

Github issue: [link](https://github.com/ionos-cloud/terraform-provider-ionoscloud/issues/144)

## Checklist

- [X] Tests added or updated
- [X] Documentation updated
- [X] Changelog updated and version incremented
- [X] Github Issue linked if any

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 5.2.20
-
-- **bug fixes**: revert icmp_code and icmp_type to string to allow setting to 0
-
 ## 5.2.19
 
 - **code enhancements**: added http request time log for API calls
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.2.20
```
IuliaBlindu, line 1: you don t have to increment the version because we did not release 5.2.19

docs/resources/firewall.md
```diff
@@ -29,15 +29,15 @@ resource "ionoscloud_firewall" "example" {
 * `datacenter_id` - (Required)[string] The Virtual Data Center ID.
 * `server_id` - (Required)[string] The Server ID.
 * `nic_id` - (Required)[string] The NIC ID.
-* `protocol` - (Required)[string] The protocol for the rule: TCP, UDP, ICMP, ANY. This property is immutable.
+* `protocol` - (Required)[string] The protocol for the rule. Property cannot be modified after creation (disallowed in update requests).
```
IuliaBlindu, line 32: Why did you removed the accepted values?

IuliaBlindu (approved):

PR #172, opened 2021-12-24
fix: primary_ip should be set only for resource

cristiGuranIonos:

primary_ip should not be set for server data source.
Primary_ip should be set when importing server resources.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -592,12 +592,14 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, meta inte
 			diags := diag.FromErr(fmt.Errorf("error while setting primary nic %s: %w", d.Id(), err))
 			return diags
 		}
-	}
-
-	if nic.Properties != nil && nic.Properties.Ips != nil && len(*nic.Properties.Ips) > 0 {
-		if err := d.Set("primary_ip", (*nic.Properties.Ips)[0]); err != nil {
-			diags := diag.FromErr(fmt.Errorf("error while setting primary ip %s: %w", d.Id(), err))
-			return diags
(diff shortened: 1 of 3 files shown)


ionoscloud/resource_server.go
```diff
@@ -592,6 +592,15 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, meta inte
 			diags := diag.FromErr(fmt.Errorf("error while setting primary nic %s: %w", d.Id(), err))
 			return diags
 		}
+		if (*server.Entities.Nics.Items)[0].Properties != nil &&
```
gabisavu98, line 595: if you want to avoid the chances of bugs occurring during a refactor or when you write the code, you can store the code sequence: *server.Entities.Nics.Items)[0].Properties.Ips  in a variable(for example: serverIps).
Plus this is a good practice because you avoid spaghetti code too so the code is easier to read.

ionoscloud/resource_server.go
```diff
@@ -866,6 +874,15 @@ func resourceServerImport(ctx context.Context, d *schema.ResourceData, meta inte
 		return nil, fmt.Errorf("error occured while fetching a server ID %s %w", d.Id(), err)
 	}
 
+	if server.Entities != nil && server.Entities.Nics != nil && (*server.Entities.Nics.Items)[0].Properties != nil &&
+		(*server.Entities.Nics.Items)[0].Properties.Ips != nil &&
```
gabisavu98, line 878: same comment as above(line 595)

gabisavu98 (approved):

PR #199, opened 2022-02-04
fix: increase nodepool timeout

cristiGuranIonos:

Increase timeout for nodepools.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [X] Changelog updated and version incremented (label: upcoming release)
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/provider.go
+++ b/ionoscloud/provider.go
@@ -256,3 +256,12 @@ var resourceDefaultTimeouts = schema.ResourceTimeout{
 	Delete:  schema.DefaultTimeout(60 * time.Minute),
 	Default: schema.DefaultTimeout(60 * time.Minute),
 }
+
+// k8sTimeouts sets default value for each Timeout type
+// can be increased to a maximum of 4 hours
+var k8sTimeouts = schema.ResourceTimeout{
+	Create:  schema.DefaultTimeout(2 * time.Hour),
+	Update:  schema.DefaultTimeout(2 * time.Hour),
+	Delete:  schema.DefaultTimeout(2 * time.Hour),
+	Default: schema.DefaultTimeout(2 * time.Hour),
+}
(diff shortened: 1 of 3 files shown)


ionoscloud/provider.go
```diff
@@ -256,3 +256,12 @@ var resourceDefaultTimeouts = schema.ResourceTimeout{
 	Delete:  schema.DefaultTimeout(60 * time.Minute),
 	Default: schema.DefaultTimeout(60 * time.Minute),
 }
+
+// k8sTimeouts sets default value for each Timeout type
+// can be increased to a maximum of 4 hours
+var k8sTimeouts = schema.ResourceTimeout{
```
gabisavu98, line 262: do we need k8sTimeouts? or we can use resourceDefaultTimeouts? 
Is resourceDefaultTimeouts used somwhere else?

ionoscloud/provider.go
```diff
@@ -256,3 +256,12 @@ var resourceDefaultTimeouts = schema.ResourceTimeout{
...
+	Create:  schema.DefaultTimeout(2 * time.Hour),
```
gabisavu98, line 263: Please extract the number 60 in a constant to be more clearly.
eg: resourceDefaultTimeoutsMinutes

gabisavu98 (commented):
Please sort imports

ionoscloud/provider.go
cristiGuranIonos: used for all other resources. k8s nodepools can take longer than any other type of resources, so we want the timeout to be separate.

ionoscloud/provider.go
cristiGuranIonos: I'd argue that it just adds a level of indirection and that 60* time.Minute is pretty clear considering resourceDefaultTimeouts is the name of the variable.

gabisavu98 (approved):

PR #224, opened 2022-03-10
fix: create utils package

cristiGuranIonos:

Creates utils package for the entire project.

## Checklist

- [] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


ionoscloud/utils.go
```diff
@@ -196,18 +152,6 @@ func responseBody(resp *ionoscloud.APIResponse) string {
...
-	if rVal.Kind() == reflect.Ptr {
-		if !rVal.IsNil() {
-			m[prop] = rVal.Elem().Interface()
-		}
-	} else {
-		m[prop] = v
-	}
-}
-
 func IsValidUUID(uuid string) bool {
```
IuliaBlindu, line 155: Is this used now, anywhere? Since we have validation on uuid from terraform

IuliaBlindu (approved):

ionoscloud/utils.go
cristiGuranIonos: Good point, some volume and server fields have this check, but we should probably replace with the terraform validation.

cristiGuranIonos: After looking a little more, we do use the function internally to check if a field is in UUID format.

PR #283, opened 2022-05-20
fix: fixed crash when trying to disable autoscaling on k8s

iblindu:

- introduced error when trying to set `max_node_count` equal to `min_node_count` in `k8s_node_pool`
- fixed crash when trying to disable `autoscaling` on `k8s_node_pool`

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [x] Jira task updated

### DIFF
--- a/ionoscloud/resource_k8s_node_pool.go
+++ b/ionoscloud/resource_k8s_node_pool.go
@@ -390,6 +390,32 @@ func getLanResourceData(lansList *schema.Set) []ionoscloud.KubernetesNodePoolLan
 	}
 	return lans
 }
+
+func getAutoscalingData(d *schema.ResourceData) (*ionoscloud.KubernetesAutoScaling, error) {
+	var autoscaling ionoscloud.KubernetesAutoScaling
+
+	asmnVal, asmnOk := d.GetOk("auto_scaling.0.min_node_count")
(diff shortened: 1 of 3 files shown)


ionoscloud/resource_k8s_node_pool.go
```diff
@@ -468,6 +460,18 @@ func resourcek8sNodePoolCreate(ctx context.Context, d *schema.ResourceData, meta
 		k8sNodepool.Properties.MaintenanceWindow.DayOfTheWeek = &mdVal
 	}
 
+	if autoscaling, err := getAutoscalingData(d); err != nil {
+		return diag.FromErr(err)
+	} else {
```
cristiGuranIonos, line 465: probably don't need else?

cristiGuranIonos (approved):

ionoscloud/resource_k8s_node_pool.go
iblindu: i can use autoscaling just in the if-else statement since it's not seen outside of it

iblindu: i can use autoscaling just in the if-else statement since it's not seen outside of it

PR #290, opened 2022-07-04
doc: use https instead of ssh for cloning

avirtopeanu-ionos:
HTTPS allows read access to the repository without the need to authenticate

Cloning via SSH requires users to set up their public ssh key on Github, which might be time consuming if they haven't set up a SSH key yet. If they simply want to use Terraform, they can just clone using HTTPS which doesn't ask for any credentials for cloning/pulling (i.e. "read" operations)

## Checklist

- [x] Documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,4 +127,4 @@ by doing the following:
 
 Set env variable IONOS_PINNED_CERT=<insert_sha256_public_fingerprint_here>
 
-You can get the sha256 fingerprint most easily form the browser by inspecting the certificate.
+You can get the sha256 fingerprint most easily from the browser by inspecting the certificate.


README.md
```diff
@@ -127,4 +127,4 @@ by doing the following:
 
 Set env variable IONOS_PINNED_CERT=<insert_sha256_public_fingerprint_here>
 
-You can get the sha256 fingerprint most easily form the browser by inspecting the certificate.
\ No newline at end of file
+You can get the sha256 fingerprint most easily form the browser by inspecting the certificate.
```
cristiGuranIonos, line 130: I just now see that I wrote 'form' instead of 'from'. I know it's not from you, but can you change that too please?

cristiGuranIonos (approved):

PR #335, opened 2022-11-11
fix: spaces, add fmt step to build workflow

cristiGuranIonos:

Run fmt every workflow.
Fix existent format issues.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,9 +21,8 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       - name: Check code is formatted properly
         run: |
-          UNFORMATTED_FILES=$(find . -not -path "./vendor/*" -name "*.go" | xargs gofmt -s -l)
-          if [ "$(UNFORMATTED_FILES | wc -l)" -gt 0 ]; then
+          if [ "$(find . -not -path "./vendor/*" -name "*.go" | xargs gofmt -s -l | wc -l)" -gt 0 ]; then
             echo "Found code in need of formatting, please run 'gofmt -s -w' before creating the PR on the files below"
-            gofmt -l -s . | grep -v vendor
+            find . -not -path "./vendor/*" -name "*.go" | xargs gofmt -s -l
             exit 1;
           fi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,9 +20,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       - name: Check code is formatted properly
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -19,3 +19,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+      - name: Check code is formatted
+        run: |
+          if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
+            echo "Found code needing formating, please run gofmt -s -w before creating the PR"
```
adeatcu-ionos, line 25: @cristiGuranIonos not sure if it matters but when we are doing the check in the if condition, we are using `-l` option, but, in the displayed message, we are asking the user to run `go fmt` using `'-s` and `-w` options. I just want to make sure that we use proper options both in check / message displayed to the user.

ionoscloud/resource_s3_key_test.go
```diff
@@ -128,20 +128,21 @@ resource ` + S3KeyResource + ` ` + S3KeyTestResource + ` {
...
-//  active     = true
-//}`
+//	resource ` + UserResource + ` "example" {
+//	 first_name = "terraform"
+//	 last_name = "test"
+//	 email = "` + utils.GenerateEmail() + `"
+//	 password = "abc123-321CBA"
+//	 administrator = false
+//	 force_sec_auth= false
+//	}
```
adeatcu-ionos, line 140: @cristiGuranIonos again commented code but I'm getting used to it. I know that these changes are just because you formatted and are only indentation-related but we really need to take a look in the future and see what we will do with this commented code.

.github/workflows/build.yml
adeatcu-ionos: @cristiGuranIonos explained me: `-l` option is used for listing, `-w` option is used for formatting, so it is correct.

.github/workflows/build.yml
```diff
@@ -19,3 +19,23 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+  gofmt:
+    name: "Run gofmt check"
+    runs-on: ubuntu-latest
+    steps:
+      -
```
adeatcu-ionos, line 26: @cristiGuranIonos not sure if this affects the functionality but shouldn't be `- name`, I mean the dash shouldn't be before `name` keyword?

cristiGuranIonos: it can be on the previous line also. moved all `-` to same line

adeatcu-ionos (approved):

PR #338, opened 2022-11-17
Feat/add ssh keys

gabisavu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


ionoscloud/resource_server.go
```diff
@@ -673,7 +717,7 @@ func SetVolumeProperties(volume ionoscloud.Volume) map[string]interface{} {
 			for _, sshKey := range *volume.Properties.SshKeys {
 				sshKeys = append(sshKeys, sshKey)
 			}
-			volumeMap["ssh_keys"] = sshKeys
+			volumeMap["ssh_keys"] = sshKeys //////////////////////////////////////////////////////
```
adeatcu-ionos, line 720: @gabisavu-ionos please remove the slashes.

ionoscloud/resource_volume.go
```diff
@@ -650,6 +655,31 @@ func getVolumeData(d *schema.ResourceData, path string) (*ionoscloud.VolumePrope
...
+	}
+	//else if v, ok := d.GetOk("ssh_keys"); ok { AR VENI PRIMUL ELSE IF MAI SUS
+	//	sshkeyPath = v.([]interface{})
+	//	if err := d.Set("ssh_key_path", v.([]interface{})); err != nil {
+	//		return nil, err
+	//	}
+	//}
+
+	if v, ok := d.GetOk(path + "ssh_keys"); ok {
+		sshkeyPath = v.([]interface{})
```
adeatcu-ionos, line 670: @gabisavu-ionos again, here we are referring to `ssh_keys`, so please change the name of the variable. Please make this change in every place in this PR where we are referring to `ssh_keys`.

ionoscloud/resource_volume.go
```diff
@@ -650,6 +655,31 @@ func getVolumeData(d *schema.ResourceData, path string) (*ionoscloud.VolumePrope
...
+		if err := d.Set("ssh_keys", v.([]interface{})); err != nil {
```
adeatcu-ionos, line 671: @gabisavu-ionos first, you cast and then store the result in `sshkeyPath`, then, in the condition, you cast again, can't we just use the previously casted and stored result?

ionoscloud/resource_volume.go
```diff
@@ -650,6 +655,31 @@ func getVolumeData(d *schema.ResourceData, path string) (*ionoscloud.VolumePrope
...
+			return nil, err
+		}
+	} else if v, ok := d.GetOk("ssh_keys"); ok {
+		sshkeyPath = v.([]interface{})
+		if err := d.Set("ssh_keys", v.([]interface{})); err != nil {
```
adeatcu-ionos, line 676: @gabisavu-ionos same here, can't we do something like: `if err := d.Set("ssh_keys", sshKey)`? Note that, for the example, I renamed `sshkeyPath` to `sshKey`.

ionoscloud/resource_volume.go
```diff
@@ -65,6 +65,11 @@ func resourceVolume() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Optional: true,
 			},
+			"ssh_keys": {
```
adeatcu-ionos, line 68: @gabisavu-ionos please add some tests in `resource_volume_test.go` that reflect the changes made in `resource_volume.go`.

adeatcu-ionos (changes requested):

ionoscloud/resource_server.go
```diff
@@ -160,15 +170,48 @@ func resourceServer() *schema.Resource {
 							Optional:   true,
 							Deprecated: "Please use ssh_key_path under server level",
 							Computed:   true,
+							ForceNew:   true,
 							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
gabisavu-ionos: I've tested creating a server with a volume which has ssh_key_path and it's fine, so there is no need of DiffSuppressFunc on volume level. We want to use DiffSuppressFunc only when we have a field that trying to change in plan before we want to make terraform apply, but we know this change will not affect the infrastructure described in the plan.

docs/resources/server.md
```diff
@@ -143,12 +143,14 @@ resource "ionoscloud_server" "example" {
 - `primary_ip` - (Computed) The associated IP address.
 - `firewallrule_id` - (Computed) The associated firewall rule.
 - `ssh_key_path` - (Optional)[list] List of paths to files containing a public SSH key that will be injected into IonosCloud provided Linux images.  Also accepts ssh keys directly. Required for IonosCloud Linux images. Required if `image_password` is not provided.
```
cristiGuranIonos, line 145: add the immutable comment here please

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: @gabisavu-ionos not sure what you mean by `"I've tested creating a server with a volume which has ssh_key_path and it's fine"`, I didn't mention something about functionality (I assumed already that you tested and it's working). 

> so there is no need of DiffSuppressFunc on volume level

I was speaking about adding this `DiffSuppressFunc` for **server's** `ssh_keys` and `ssh_key_path`, not on volume level.

> only when we have a field that trying to change in plan before we want to make terraform apply, but we know this change will not affect the infrastructure described in the plan.

I didn't understand this part, can you please rephrase? 

And, to be more clear, when I wrote this comment I did it because I noticed that we have `DiffSuppressFunc` for `ssh_keys` and `ssh_key_path`, so I assumed that, for these fields, we need to have this function (regardless of where the fields are located: on volume level or on server level). Right now, if I provide `ssh_keys` in the tf plan on the **volume** resource within a server resource,  then `DiffSuppressFunc` is called and an extra check is done inside this function, but if I provide `ssh_keys` only on the **server** resource, then this function is **not** called and the check is **not** done and I'm not sure if we want this. It made more sense to me to use the `DiffSuppressFunc` in both cases, since we are talking about the same fields: `ssh_keys`/`ssh_key_path`.

ionoscloud/resource_server.go
```diff
@@ -170,7 +181,31 @@ func resourceServer() *schema.Resource {
...
-								if len(utils.DiffSlice(convertSlice(sshKeyPath), convertSlice(oldSshKeyPath))) == 0 {
+								difKeypath := utils.DiffSlice(convertSlice(sshKeyPath), convertSlice(oldSshKeyPath))
+								if len(difKeypath) == 0 {
+									return true
+								}
+
+								return false
+							},
+						},
+						"ssh_keys": {
```
cristiGuranIonos, line 192: add the description here too please

ionoscloud/test_constants.go
```diff
@@ -153,6 +153,117 @@ resource ` + ServerResource + ` ` + ServerTestResource + ` {
 }`
 )
 
+const (
```
cristiGuranIonos, line 156: no need for const (, just defined const testAccCheckServerSshKeysDirectly

ionoscloud/test_constants.go
```diff
@@ -153,6 +153,117 @@ resource ` + ServerResource + ` ` + ServerTestResource + ` {
...
+	  source_mac = "00:0a:95:9d:68:17"
+	  source_ip = ionoscloud_ipblock.webserver_ipblock.ips[2]
+	  target_ip = ionoscloud_ipblock.webserver_ipblock.ips[3]
+	  type = "EGRESS"
+    }
+  }
+}`
+)
+
+const (
```
cristiGuranIonos, line 211: no need for const (, just defined const testAccCheckServerSshKeysAndKeyPath

ionoscloud/resource_volume.go
adeatcu-ionos: I will mark this as solved since we will remove `ssh_keys` and `ssh_key_path` for the volume resource. If we were to add tests now, it would mean deleting them shortly after, so it' better not to add them at all.

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: @gabisavu-ionos I mark this as `unsolved` again, let's have a quick call when you can to discuss the details.

ionoscloud/resource_server.go
```diff
@@ -170,7 +181,31 @@ func resourceServer() *schema.Resource {
...
+							Type:       schema.TypeList,
+							Elem:       &schema.Schema{Type: schema.TypeString},
+							Optional:   true,
+							Deprecated: "Please use ssh_keys under server level",
+							Computed:   true,
```
adeatcu-ionos, line 197: @gabisavu-ionos please also add `ForceNew:   true` here.

adeatcu-ionos (changes requested):

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: Marking this as solved again, I discussed with  @gabisavu-ionos and we decided to leave this as it is since the `DiffSuppressFunc` will be removed in the future. Besides that, right now, the function is not used properly, since it is more of a workaround to see if the `ssh_keys` order is the same for the keys set on `volume` and `server` levels.

PR #346, opened 2022-12-13
fix: get nic id from entities and %w fix in printf

cristiGuranIonos:

Get nic id from first element in entities, if lan 1 not found.
Fix %w in printf

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -1350,16 +1350,10 @@ func setResourceServerData(ctx context.Context, client *ionoscloud.APIClient, d 
 	var nicId string
 	if primaryNicOk {
 		nicId = d.Get("primary_nic").(string)
-	} else if server.Entities.Nics != nil && server.Entities.Nics.Items != nil && len(*server.Entities.Nics.Items) > 0 && (*server.Entities.Nics.Items)[0].Id != nil { // this might be a terraformer import, so primary_nic might not be set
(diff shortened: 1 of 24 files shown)


ionoscloud/resource_server.go
```diff
@@ -1356,6 +1356,10 @@ func setResourceServerData(ctx context.Context, client *ionoscloud.APIClient, d
 				nicId = *nic.Id
```
adeatcu-ionos, line 1356: @cristiGuranIonos looking at this again I wonder if we can just get rid of this check, because, basically, in this check, we are setting the `nicId` as being the `ID` of the last `NIC` in the list of `NICs` that has `LAN` 1. So instead of doing this, I think we can do:
```
nicId = *(*server.Entities.Nics.Items)[0].Id
```
directly since we already know that the list contains at least one `NIC` and the `ID` of the first `NIC` is not `nil`.

More than that, we decided that we should populate this with the FIRST `NIC` from the list, but with this code I'm not sure if this happens all the time, maybe there are some cases in which the `NIC` that is in the first `LAN` is not the first in the list. What do you think?

cristiGuranIonos: I think we should just remove the for that checks for lan 1 and just set the first id we find in the list

adeatcu-ionos (approved):

gabisavu-ionos (approved):

Issue #379, opened 2023-03-30
Deleting Server with terraform does not delete volume


adnkai:
## Description
Deleting a Server with terraform results in inconsistent tfstate as the volume is not deleted in IONOS cloud.

## Expected behavior
Volume should be deleted or at least described in tfstate.

## Environment

Terraform version: 
```
1.4.2
```
Provider version: 
```
6.3.6
```
OS:
```
Windows
```

## Example
```
terraform {
    required_providers {
        ionoscloud = {
            source = "ionos-cloud/ionoscloud"
            version = "= 6.3.6"
        }
    }
}

provider "ionoscloud" {
    token = <REDACTED>
    endpoint = "https://api.ionos.com/cloudapi/v6/"
}

resource "ionoscloud_server" "SBlade01" {
    availability_zone = "AUTO"
    cores             = 4
    cpu_family        = "INTEL_SKYLAKE"
    datacenter_id     = <REDACTED>
    name              = "SBlade01"
    ram               = 8192
    type              = "ENTERPRISE"

    nic {
        dhcp            = true
        firewall_active = true
        firewall_type   = "INGRESS"
        lan             = 1
    }

    volume {
        availability_zone      = "AUTO"
        bus                    = "VIRTIO"
        disk_type              = "SSD Standard"
        licence_type           = "UNKNOWN"
        name                   = "SBlade01"
        size                   = 50
    }
}
```

## How to Reproduce

1. Create main.tf with resources (ionoscloud_server with volume)
2. remove ionoscloud_server from main.tf
3. server is being deleted with terraform. Volume persists in IONOS

cristiGuranIonos:
Thanks for creating this. Could not reproduce, for me the volume is deleted. Some questions:
- How often does it reproduce? 
- Can you upload some logs if it reproduces with TF_LOG set to debug? Was there any error thrown? 
- How do you check the volume is still there? Do you use ionosctl?
- What type of image do you use on the server?
Thanks!

adnkai:
Hey! Thanks for the quick response!
We're not using any specific image. What you see in my post is the entire main.tf.
It happens every time. Creating the resources in a new, empty DCD. 
Removing resources from main.tf to have Terraform delete them. 
The volume persists, dangling around all on it's own.
Happens by running `terraform apply` from my local machine and from Terraform.io.
We don't check it with ionosctl, we just see the volume it in the webUI.

This does not happen with cubes, just servers.

[...]

PR #380, opened 2023-04-04
fix: delete volumes from a serer when no boot volume id is provided

gabi-ionos:
Solves issue #379 

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -164,6 +164,7 @@ resource "random_password" "server_image_password" {
 - `image_password` - (Optional)[string] Required if `ssh_key_path` is not provided.
 - `type` - (Optional)[string] Server usages: [ENTERPRISE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/virtual-servers) or [CUBE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/cloud-cubes). This property is immutable.
 - `label` - (Optional) A label can be seen as an object with only two required fields: `key` and `value`, both of the `string` type. Please check the example presented above to see how a `label` can be used in the plan. A server can have multiple labels.
+- `delete_volumes` - (Optional) [bool] deleting all the volumes of a server even if no boot_volume provided
 
 > **⚠ WARNING** 
 > 
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -164,7 +164,7 @@ resource "random_password" "server_image_password" {
 - `image_password` - (Optional)[string] Required if `ssh_key_path` is not provided.
 - `type` - (Optional)[string] Server usages: [ENTERPRISE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/virtual-servers) or [CUBE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/cloud-cubes). This property is immutable.
 - `label` - (Optional) A label can be seen as an object with only two required fields: `key` and `value`, both of the `string` type. Please check the example presented above to see how a `label` can be used in the plan. A server can have multiple labels.
-- `delete_volumes` - (Optional) [bool] deleting all the volumes of a server even if no boot_volume provided
+- `delete_volumes` - (Optional)[bool] Enforce the deletion of all volumes of a server, not only boot_volume.
 
 > **⚠ WARNING** 
 > 
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -164,7 +164,6 @@ resource "random_password" "server_image_password" {
 - `image_password` - (Optional)[string] Required if `ssh_key_path` is not provided.
 - `type` - (Optional)[string] Server usages: [ENTERPRISE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/virtual-servers) or [CUBE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/cloud-cubes). This property is immutable.
 - `label` - (Optional) A label can be seen as an object with only two required fields: `key` and `value`, both of the `string` type. Please check the example presented above to see how a `label` can be used in the plan. A server can have multiple labels.
-- `delete_volumes` - (Optional)[bool] Enforce the deletion of all volumes of a server, not only boot_volume.
 
 > **⚠ WARNING** 
 > 
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -1057,6 +1057,7 @@ func resourceServerDelete(ctx context.Context, d *schema.ResourceData, meta inte
 	client := meta.(SdkBundle).CloudApiClient
 	dcId := d.Get("datacenter_id").(string)
 
+	// A bigger depth is required since we need all volumes items.
 	server, apiResponse, err := client.ServersApi.DatacentersServersFindById(ctx, dcId, d.Id()).Depth(2).Execute()
 	logApiRequestTime(apiResponse)
 
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -1056,12 +1056,21 @@ func resourceServerDelete(ctx context.Context, d *schema.ResourceData, meta inte
 	client := meta.(SdkBundle).CloudApiClient
 	dcId := d.Get("datacenter_id").(string)
 
-	server, apiResponse, err := client.ServersApi.DatacentersServersFindById(ctx, dcId, d.Id()).Execute()
+	server, apiResponse, err := client.ServersApi.DatacentersServersFindById(ctx, dcId, d.Id()).Depth(2).Execute()
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
 		diags := diag.FromErr(fmt.Errorf("error occured while fetching a server ID %s %w", d.Id(), err))
 		return diags
+	}
+
+	if deleteVolumes, ok := d.GetOk("delete_volumes"); ok {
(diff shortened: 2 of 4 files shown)


docs/resources/server.md
```diff
@@ -164,6 +164,7 @@ resource "random_password" "server_image_password" {
 - `image_password` - (Optional)[string] Required if `ssh_key_path` is not provided.
 - `type` - (Optional)[string] Server usages: [ENTERPRISE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/virtual-servers) or [CUBE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/cloud-cubes). This property is immutable.
 - `label` - (Optional) A label can be seen as an object with only two required fields: `key` and `value`, both of the `string` type. Please check the example presented above to see how a `label` can be used in the plan. A server can have multiple labels.
+- `delete_volumes` - (Optional) [bool] deleting all the volumes of a server even if no boot_volume provided
```
adeatcu-ionos, line 167: @gabi-ionos it shouldn't be a space between `(Optional)` and `[bool]`, it should be `(Optional)[bool]`. 

Also, I would change this description since it makes me think that if `boot_volume` is provided, all volumes will be deleted which it's false. 

I suggest something like: `(Optional)[bool] Enforce the deletion of all volumes of a server, not only boot_volume.` This also makes clear that if `delete_volumes` is `false`, only the `boot_volume` will be deleted (this happens now with our current code but I will describe what I think related to this behavior in another comment).

ionoscloud/resource_server.go
```diff
@@ -1026,6 +1031,27 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 	return resourceServerRead(ctx, d, meta)
 }
 
+func deleteAllVolumes(ctx context.Context, d *schema.ResourceData, client *ionoscloud.APIClient, server *ionoscloud.Server) diag.Diagnostics {
```
adeatcu-ionos, line 1034: @gabi-ionos not sure how you tested this, but it's not working. I created a server using the configuration presented in the [issue](https://github.com/ionos-cloud/terraform-provider-ionoscloud/issues/379) and then I manually added two volumes using the `DCD`. After this, I removed the server from the `tf` plan and, after deletion, there were dangling volumes in the `DCD`.

ionoscloud/resource_server.go
```diff
@@ -1026,6 +1031,27 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+	dcId := d.Get("datacenter_id").(string)
+
+	if entities, ok := server.GetEntitiesOk(); ok {
+		if volumes, ok := entities.GetVolumesOk(); ok {
+			if volumes.Items != nil {
```
adeatcu-ionos, line 1039: @gabi-ionos I played with the debugger a little bit, `volumes.Items` is always `nil` here, the volumes are never deleted and hence the dangling volumes from the previous comment.

ionoscloud/resource_server.go
```diff
@@ -1026,6 +1031,27 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+				for _, volume := range *volumes.Items {
+					apiResponse, err := client.VolumesApi.DatacentersVolumesDelete(ctx, dcId, *volume.GetId()).Execute()
+					logApiRequestTime(apiResponse)
+
+					if err != nil {
+						diags := diag.FromErr(fmt.Errorf("error occured while deleting remaining volume %s of server ID %s %w", *server.Properties.BootVolume.Id, d.Id(), err))
```
adeatcu-ionos, line 1045: @gabi-ionos here it shouldn't be `*server.Properties.BootVolume.Id`.

ionoscloud/resource_server_test.go
```diff
@@ -216,6 +216,45 @@ func TestAccServerBasic(t *testing.T) {
 	})
 }
 
+func TestAccServerNoBootVolumeBasic(t *testing.T) {
```
adeatcu-ionos, line 219: @gabi-ionos this test has nothing to do with [this](https://github.com/ionos-cloud/terraform-provider-ionoscloud/issues/379) issue, here you are only checking the creation of a server with a volume that it's not a boot volume (like we do for other tests), the only visible difference is that you added `delete_volumes = true` which has no effect on the creation side, so this test is useless for the issue's context.

adeatcu-ionos (changes requested):

adeatcu-ionos:
@gabi-ionos @cristiGuranIonos I analyzed the behavior for `DCD` vs `Terraform` related to server deletion:

In `DCD`, after you delete a server, you are asked if you want to delete all the volumes, if you press yes, all volumes are deleted, if you press no, no volume is deleted.

In `Terraform`, with the current implementation, if we specify `delete_volumes = true` in the `tf` plan, all the volumes are deleted (this is not happening now but it will happen after code fixes), but if we don't specify it, only the `boot_volume`, it's deleted (if it exists, of course).

So, as you can notice, there is a small difference between the two. In my opinion, I think that the `DCD` behavior makes sense, but we have some checks, like this one: https://github.com/ionos-cloud/terraform-provider-ionoscloud/blob/master/ionoscloud/resource_server.go#L1041 and even though there are no comments related to it, the check was added there for a reason, so not sure about this.

I wrote this to make things clearer, maybe we can discuss related to the future implementation.

ionoscloud/resource_server.go
```diff
@@ -1026,18 +1031,48 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+	}
+	return nil
+}
+
 func resourceServerDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(SdkBundle).CloudApiClient
 	dcId := d.Get("datacenter_id").(string)
 
-	server, apiResponse, err := client.ServersApi.DatacentersServersFindById(ctx, dcId, d.Id()).Execute()
+	server, apiResponse, err := client.ServersApi.DatacentersServersFindById(ctx, dcId, d.Id()).Depth(2).Execute()
```
cristiGuranIonos, line 1059: comment here why we need depth(2)

ionoscloud/resource_server.go
```diff
@@ -130,6 +130,11 @@ func resourceServer() *schema.Resource {
 				ForceNew:      true,
 				Description:   "Public SSH keys are set on the image as authorized keys for appropriate SSH login to the instance using the corresponding private key. This field may only be set in creation requests. When reading, it always returns null. SSH keys are only supported if a public Linux image is used for the volume creation.",
 			},
+			"delete_volumes": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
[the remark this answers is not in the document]
cristiGuranIonos: delete_volumes needs to exist only for enterprise servers, right? We need to say that.

adeatcu-ionos:
@gabi-ionos @cristiGuranIonos leaving this here to make it transparent for the others. We had a discussion about the best approach related to deleting volumes, boot volumes, etc.

Since we can have multiple volumes attached to a server without specifying a boot volume, we decided that, for now, the best approach would be to delete all the volumes. This ensures that, after server deletion, there are no dangling volumes left and there are no conflicts that can occur related to `terraform` interaction (planning, etc.).

So, on the current PR, we need to remove `delete_volumes` attribute and modify the `resourceServerDelete` in order to delete all volumes that are attached to an `ENTERPRISE` server. Note that we need to wait for each volume removal during the deletion process.

main.go
```diff
@@ -13,7 +13,7 @@ func main() {
 
 	//set this to true to enable cli debugging your provider, by running headless and setting TF_REATTACH_PROVIDERS on the terraform terminal
 	//this will enable you to debug when running plans from cli.
-	flag.BoolVar(&debugMode, "debuggable", false, "set to true to run the provider with support for debuggers like delve/goland")
+	flag.BoolVar(&debugMode, "debuggable", true, "set to true to run the provider with support for debuggers like delve/goland")
```
cristiGuranIonos, line 16: remove this please

docs/resources/server.md
adeatcu-ionos: @cristiGuranIonos we removed `delete_volumes` attribute, so this comment is no longer available.

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos we removed `delete_volumes` attribute, so this comment is no longer available.

ionoscloud/resource_server.go
adeatcu-ionos: @cristiGuranIonos done in 46795f44335487638eb6bc12084809359ce5a660

main.go
adeatcu-ionos: @cristiGuranIonos this was removed in 6728515cf24d826bd9e51ba0349cce52bdfc8dd6

ionoscloud/resource_server_test.go
adeatcu-ionos: @cristiGuranIonos I see that the tests were modified, I also modified the function that checks if the servers/volumes are deleted properly. I did this in this commit: fa6afaad843f8bf9dbd6f1573b1fd37bc3272a79

adeatcu-ionos (approved):

cristiGuranIonos (approved):

PR #383, opened 2023-04-21
feat: add support for DNS

adeatcu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


adeatcu-ionos:
@cristiGuranIonos please check this out when you have time, this is the first iteration. During your review, I'll write the documentation and after this initial review, I'll also write the tests. Thank you!

services/dnsaas/zone.go
```diff
@@ -0,0 +1,142 @@
...
+	zones, apiResponse, err := c.sdkClient.ZonesApi.ZonesGetExecute(request)
+	apiResponse.LogInfo()
+	return zones, apiResponse, err
+}
+
+func (c *Client) SetZoneData(d *schema.ResourceData, zone dnsaas.ZoneResponse) error {
+	if zone.Id != nil {
+		d.SetId(*zone.Id)
+	}
+
```
cristiGuranIonos, line 59: if zone.properties == nil - return error

adeatcu-ionos: @cristiGuranIonos solved in 2273efbf55a163a4f7e5e4a2f4f6bbe854780969 for both zones and records. I also added documentation. If everything else is ok, I can start writing tests.

ionoscloud/data_source_dnsaas_record.go
```diff
@@ -0,0 +1,143 @@
...
+	dnsaas "github.com/ionos-cloud/sdk-go-dnsaas"
+	"log"
+	"strings"
+)
+
+func dataSourceDNSaaSRecord() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceRecordRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
```
cristiGuranIonos, line 17: maybe a UUID validation here and for zone_id

ionoscloud/data_source_dnsaas_record.go
```diff
@@ -0,0 +1,143 @@
...
+		var results []dnsaas.RecordResponse
+		log.Printf("[INFO] Populating data source for DNS Record using name: %s and partial_match: %t", recordName, partialMatch)
+		if partialMatch {
+			// By default, when providing the name as a filter, for the GET requests, partial match
+			// is true.
+			records, _, err := client.ListRecords(ctx, zoneId, recordName)
+			if err != nil {
+				return diag.FromErr(fmt.Errorf("an error occured while fetching DNS Records: %w", err))
+			}
+			if records.Items != nil {
```
cristiGuranIonos, line 101: maybe just break if records.Items == nil and the error will be handled by `if results == nil || len(results) == 0 {`

ionoscloud/resource_dnsaas_record.go
```diff
@@ -0,0 +1,175 @@
...
+
+func recordCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(SdkBundle).DNSaaSClient
+	zoneId := d.Get("zone_id").(string)
+
+	recordResponse, _, err := client.CreateRecord(ctx, zoneId, d)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("an error occured while creating a record for the DNS zone with ID: %s, error: %w", zoneId, err))
+	}
+	if recordResponse.Metadata == nil {
```
cristiGuranIonos, line 75: Do we have to wait  until it's in `created` state? Or is it synchronous?

ionoscloud/data_source_dnsaas_record.go
adeatcu-ionos: @cristiGuranIonos done in 01e2b65f9fe5d96e6c21367e7fe3b2e7b6343ba2

ionoscloud/data_source_dnsaas_record.go
adeatcu-ionos: @cristiGuranIonos done in 01e2b65f9fe5d96e6c21367e7fe3b2e7b6343ba2, I completely removed it.

ionoscloud/data_source_dnsaas_record.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		} else {
+			// In order to have an exact name match, we must retrieve all the DNS Records and then
+			// build a list of exact matches based on the response, there is no other way since using
+			// filter.name only does a partial match.
+			records, _, err := client.ListRecords(ctx, zoneId, "")
+			if err != nil {
+				return diag.FromErr(fmt.Errorf("an error occured while fetching DNS Records: %w", err))
+			}
+			if records.Items != nil {
```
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos done in 01e2b65f9fe5d96e6c21367e7fe3b2e7b6343ba2, I completely removed it.

ionoscloud/data_source_dnsaas_zone.go
```diff
@@ -0,0 +1,133 @@
...
+		log.Printf("[INFO] Populating data source for DNS Zone using name %s and partial_match %t", name, partialMatch)
+
+		if partialMatch {
+			// By default, when providing the name as a filter, for the GET requests, partial match
+			// is true.
+			zones, _, err := client.ListZones(ctx, name)
+			if err != nil {
+				return diag.FromErr(fmt.Errorf("an error occured while fetching DNS Zones: %w", err))
+			}
+			if zones.Items != nil {
```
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos done in 01e2b65f9fe5d96e6c21367e7fe3b2e7b6343ba2, I completely removed it.

cristiGuranIonos (approved):

cristiGuranIonos (approved):

cristiGuranIonos:
Add to changelog  please

ionoscloud/resource_dnsaas_record_test.go
```diff
@@ -0,0 +1,218 @@
+package ionoscloud
```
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos done in 8689a85494ef2d919fd4174d649982cce2919387

adeatcu-ionos:
> Add to changelog please

@cristiGuranIonos done in 8689a85494ef2d919fd4174d649982cce2919387

docs/data-sources/dnsaas_record.md
```diff
@@ -0,0 +1,66 @@
+---
+subcategory: "DNS as a Service"
+layout: "ionoscloud"
+page_title: "IonosCloud : ionoscloud_dns_record"
+sidebar_current: "docs-dns_record"
+description: |-
+  Get information on a DNS Record.
+---
+
+# ionoscloud_dns_zone
```
mimihalescu, line 10: ionoscloud_dns_record*

adeatcu-ionos: @mimihalescu done, thanks for catching this!

adeatcu-ionos:
@mimihalescu @cristiGuranIonos I also made the renaming, please take a look and check if I missed something, thank you!

services/dns/record.go
```diff
@@ -0,0 +1,209 @@
...
+	dns "github.com/ionos-cloud/sdk-go-dns"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/utils"
+)
+
+var recordResourceName = "DNS Record"
+
+func (c *Client) CreateRecord(ctx context.Context, zoneId string, d *schema.ResourceData) (recordResponse dns.RecordResponse, responseInfo utils.ApiResponseInfo, err error) {
+	// Must is a helper that wraps a call to a function returning (UUID, error) and panics if the error is non-nil.
+	uuidV4 := uuid.Must(uuid.NewV4())
+	recordUuid := uuid.NewV5(uuid.NewV5(uuid.NamespaceURL, "https://github.com/ionos-cloud/terraform-provider-ionoscloud"), uuidV4.String())
```
adeatcu-ionos, line 20: @cristiGuranIonos I had in mind to generate the record `UUID` also based on zone `UUID`, not sure if this adds anything useful tho, what do you think?

adeatcu-ionos:
@mimihalescu @avirtopeanu-ionos @rmocanu-ionos @cristiGuranIonos I added all of you for the review since I changed the functionality for create method, so now the creation is done via `PUT` instead of `POST`, as we discussed, please take a look and tell me what you think.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/ionos-cloud/terraform-provider-ionoscloud/v6
 go 1.20
 
 require (
+	github.com/gofrs/uuid v4.4.0+incompatible
```
cristiGuranIonos, line 6: can we use v5 instead?

services/dns/record.go
cristiGuranIonos: I don't think it's necessary

services/dns/record.go
```diff
@@ -14,8 +15,11 @@ import (
 var recordResourceName = "DNS Record"
 
 func (c *Client) CreateRecord(ctx context.Context, zoneId string, d *schema.ResourceData) (recordResponse dns.RecordResponse, responseInfo utils.ApiResponseInfo, err error) {
-	request := setRecordCreateRequest(d)
-	recordResponse, apiResponse, err := c.sdkClient.RecordsApi.ZonesRecordsPost(ctx, zoneId).RecordCreateRequest(*request).Execute()
+	// Must is a helper that wraps a call to a function returning (UUID, error) and panics if the error is non-nil.
+	uuidV4 := uuid.Must(uuid.NewV4())
+	recordUuid := uuid.NewV5(uuid.NewV5(uuid.NamespaceURL, "https://github.com/ionos-cloud/terraform-provider-ionoscloud"), uuidV4.String())
```
cristiGuranIonos, line 20: I think we can extract this part as a constant? - uuid.NewV5(uuid.NamespaceURL, "https://github.com/ionos-cloud/terraform-provider-ionoscloud")

services/dns/zone.go
```diff
@@ -14,8 +15,11 @@ import (
 var zoneResourceName = "DNS Zone"
 
 func (c *Client) CreateZone(ctx context.Context, d *schema.ResourceData) (zoneResponse dns.ZoneResponse, responseInfo utils.ApiResponseInfo, err error) {
-	request := setZoneCreateRequest(d)
-	responseData, apiResponse, err := c.sdkClient.ZonesApi.ZonesPost(ctx).ZoneCreateRequest(*request).Execute()
+	// Must is a helper that wraps a call to a function returning (UUID, error) and panics if the error is non-nil.
+	uuidV4 := uuid.Must(uuid.NewV4())
+	zoneUuid := uuid.NewV5(uuid.NewV5(uuid.NamespaceURL, "https://github.com/ionos-cloud/terraform-provider-ionoscloud"), uuidV4.String())
```
avirtopeanu-ionos, line 20: in my opinion writing `uuid.NewV5(uuid.NewV5(uuid.NamespaceURL, "https://github.com/ionos-cloud/terraform-provider-ionoscloud"), uuidV4.String())` every time you want a new ID is a bit repetitive

What do you feel about creating a very small pkg `internal/tfid/tfid.go` which has the sole responsibility of just creating V5 UUIDs with the namespaceURL already set ? And you could call it like `tfid.New()` for example

vendor/modules.txt
```diff
@@ -9,6 +9,9 @@ github.com/apparentlymart/go-textseg/v13/textseg
 github.com/fatih/color
 # github.com/go-test/deep v1.0.6
 ## explicit; go 1.13
+# github.com/gofrs/uuid v4.4.0+incompatible
```
avirtopeanu-ionos, line 12: Any reason for using this lib ?

Google/uuid package is already in this go.mod and having 2 dependencies on different uuid pkgs isnt good

And the reason I prefer keeping `google/uuid` over refactoring to only use the new one is because any dev which just quickly wants an UUID V4 would default to `google/uuid` without thinking that we're actually using `gofrs/uuid`

avirtopeanu-ionos: I looked over docs and seems like `google/uuid` cant generate V5 UUIDs.

In this case I'm okay with having two different libs for UUID handling, but we should get rid of `google/uuid` as soon as possible and replace all those calls with `gofrs/uuid` equivalent

adeatcu-ionos: @avirtopeanu-ionos ` but we should get rid of google/uuid as soon as possible and replace all those calls with gofrs/uuid equivalent` I agree, I'll open an issue for this.

services/dns/zone.go
adeatcu-ionos: @avirtopeanu-ionos sure, not sure why I didn't do that in the beginning.

go.mod
adeatcu-ionos: @cristiGuranIonos done in 5369bab15a6a1015bc488c1140a42af0283ed025.

services/dns/zone.go
adeatcu-ionos: @avirtopeanu-ionos done in 5369bab15a6a1015bc488c1140a42af0283ed025.

mimihalescu (dismissed):

ionoscloud/resource_dns_zone.go
```diff
@@ -0,0 +1,147 @@
...
+	zoneId := d.Id()
+
+	zoneResponse, _, err := client.UpdateZone(ctx, zoneId, d)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("an error occured while updating the DNS Zone with ID: %s, error: %w", zoneId, err))
+	}
+	if zoneResponse.Metadata.State != nil {
+		if *zoneResponse.Metadata.State == dns.FAILED {
+			// This is a temporary error message since right now the API is not returning errors that we can work with.
+			return diag.FromErr(fmt.Errorf("zone update has failed, this can happen if the data in the request is not correct, " +
```
avirtopeanu-ionos, line 101: In this case, the API zone is still going to remain in FAILED state, right? Does throwing an error not kind of hide the fact that something on the cloud has changed after running the plan? I'm not experienced with terraform, just trying to understand what is happening here

ionoscloud/resource_dns_zone.go
```diff
@@ -0,0 +1,147 @@
...
+		if apiResponse.HttpNotFound() {
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(fmt.Errorf("error while deleting DNS Zone with ID: %s, error: %w", zoneId, err))
+	}
+
+	err = utils.WaitForResourceToBeDeleted(ctx, d, client.IsZoneDeleted)
+	if err != nil {
+		diag.FromErr(fmt.Errorf("an error occured while waiting for the DNS Zone with ID: %s to be deleted, error: %w", zoneId, err))
```
avirtopeanu-ionos, line 123: shouldn't `diag.FromErr` always be returned in case of err?

services/dns/record.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	if record.Metadata.Fqdn != nil {
+		if err := d.Set("fqdn", *record.Metadata.Fqdn); err != nil {
+			return utils.GenerateSetError(recordResourceName, "fqdn", err)
+		}
+	}
+
+	return nil
+}
```
avirtopeanu-ionos, line 106: Consider github.com/fatih/structs for dynamic field estting

ionoscloud/resource_dns_zone.go
adeatcu-ionos: `I'm not experienced with terraform` no worries

I understand your point and it makes sense, it can actually be an edge case that we I didn't think about, I will check to see the current behavior. I expect something like:

1) We try to create the Zone, we send the request and the zone is created in a `FAILED` state;
2) Because of that, we raise an error, and we are not setting anything in the `Terraform` state, this means that Terraform will not track anything related to this resource;
3) The resource will continue to exist in a `FAILED` state and basically this is not good since Terraform won't track this resource.

But I'll come back with updates.

avirtopeanu-ionos: Ok so I did a bit more digging and (please correct any of my points if wrong):
- This is known as "terraform drift" and should be corrected externally (via CLI for example)
- Throwing an error in this case is good, since the terraform plan wasn't successful
- maybe consider updating the error message to let the user know:
     1. that a zone has been created/updated/etc., however it is in state `FAILED`
     2. he needs to take some external action to revert the tf plan's changes

adeatcu-ionos: @avirtopeanu-ionos this is not really a drift since we are not setting anything in the terraform state. Usually, after you receive a response, you set some values in the state in order to have a mapping between the resource defined in the plan and the actual resurce that exists in the cloud. Also, you can check this:
```
Drift is the term for when the real-world state of your infrastructure differs from the state defined in your configuration. 
```
`Throwing an error in this case is good, since the terraform plan wasn't successful` indeed it is, but my problem is that the resource still exists in cloud, at least this is what I expect.
`maybe consider updating the error message to let the user know: ` this is a possible solution, we will check all possible solutions after I test the behavior. I can't test it RN because there are some problems with the DNS API.

ionoscloud/resource_dns_zone.go
adeatcu-ionos: @avirtopeanu-ionos done, I missed this.

ionoscloud/resource_dns_zone.go
adeatcu-ionos: @avirtopeanu-ionos I was able to test this again, but it seems like the behavior is different, I tried to create a record in the `FAILED` state, but right now I receive a 400 error response, it seems like they changed the behavior. Let's leave it as it is for now and I will ask them how I can generate a response in which the state is `FAILED` in order to check if a resource is created or not.

avirtopeanu-ionos (approved):

cristiGuranIonos (approved):

PR #384, opened 2023-05-04
fix: allow k8s_version patch upgrades, but ignore downgrades

cristiGuranIonos:

Ignore downgrades of `k8s_version` patch level.
Allow upgrades of `k8s_version` patch level.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_k8s_cluster.go
+++ b/ionoscloud/resource_k8s_cluster.go
@@ -33,7 +33,7 @@ func resourcek8sCluster() *schema.Resource {
 			},
 			"k8s_version": {
 				Type:             schema.TypeString,
-				Description:      "The desired Kubernetes Version. For supported values, please check the API documentation. Downgrades are not supported. Downgrades are not supported. The provider will ignore downgrades of patch level.",
+				Description:      "The desired Kubernetes Version. For supported values, please check the API documentation. Downgrades are not supported. The provider will ignore downgrades of patch level.",
 				Optional:         true,
 				Computed:         true,
 				DiffSuppressFunc: DiffBasedOnVersion,
--- a/ionoscloud/resource_k8s_cluster.go
+++ b/ionoscloud/resource_k8s_cluster.go
@@ -33,7 +33,7 @@ func resourcek8sCluster() *schema.Resource {
 			},
 			"k8s_version": {
 				Type:             schema.TypeString,
-				Description:      "The desired kubernetes version",
+				Description:      "The desired Kubernetes Version. For supported values, please check the API documentation. Downgrades are not supported. Downgrades are not supported. The provider will ignore downgrades of patch level.",
 				Optional:         true,
 				Computed:         true,
 				DiffSuppressFunc: DiffBasedOnVersion,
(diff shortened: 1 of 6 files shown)


ionoscloud/resource_k8s_cluster.go
```diff
@@ -33,7 +33,7 @@ func resourcek8sCluster() *schema.Resource {
 			},
 			"k8s_version": {
 				Type:             schema.TypeString,
-				Description:      "The desired kubernetes version",
+				Description:      "The desired Kubernetes Version. For supported values, please check the API documentation. Downgrades are not supported. Downgrades are not supported. The provider will ignore downgrades of patch level.",
```
adeatcu-ionos, line 36: @cristiGuranIonos `Downgrades are not supported` is written twice here.

ionoscloud/utils.go
```diff
@@ -29,24 +30,33 @@ func responseBody(resp *ionoscloud.APIResponse) string {
...
 // DiffBasedOnVersion used for k8 node pool and cluster
+// ignores downgrades of the patch versions.
 func DiffBasedOnVersion(_, old, new string, _ *schema.ResourceData) bool {
 	var oldMajor, oldMinor string
 	var newMajor, newMinor string
+	var oldPatchInt, newPatchInt int
 	if old != "" {
 		oldSplit := strings.Split(old, ".")
-		if len(oldSplit) > 1 {
+		if len(oldSplit) > 2 {
```
adeatcu-ionos, line 40: @cristiGuranIonos not sure why we need this length check since the version should always have the format `x.y.z`, meaning that the length will always be 3.

Anyway, now we have `> 2` instead of `> 1`, so we are restricting the number of cases in which we enter the condition. There are two things unclear for me:

1. Why do we need a length check?
2. Why did we increase the length in the check since we have the same format for version?

ionoscloud/utils.go
```diff
@@ -29,24 +30,33 @@ func responseBody(resp *ionoscloud.APIResponse) string {
...
 			newMinor = newSplit[1]
+			newPatchInt, _ = strconv.Atoi(newSplit[2])
 		}
 
 		if oldMajor == newMajor && oldMinor == newMinor {
-			return true
+			//this is a downgrade of the patch version that we will ignore
+			// it may happen either manually, or after a maintenance window
+			if oldPatchInt > newPatchInt {
+				log.Printf("[WARN] Downgrade is not supported on k8s from %d to %d", oldPatchInt, newPatchInt)
```
adeatcu-ionos, line 57: @cristiGuranIonos I haven't seen `WARN` in our code, only `WARNING`, let's keep it that way from now on, so please modify this.

adeatcu-ionos (changes requested):

ionoscloud/utils.go
cristiGuranIonos: To make sure that the split doesn't panic and we don't read above the array boundary.  Teoretically a user might do something like 1.20 although not supported

ionoscloud/utils.go
cristiGuranIonos: oddly enough these are actually 2 bugs in our provider that I will fix in a separate PR. Will detail there.

adeatcu-ionos (approved):

PR #391, opened 2023-05-31
feat: make nic optional and allow multiple firewalls in server in ionoscloud_server

cristiGuranIonos:

Make `nic` optional in `ionoscloud_server`.
Allow multiple `firewall` rules to be defined inside `ionoscloud_server`.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


avirtopeanu-ionos (approved):

ionoscloud/resource_server.go
```diff
@@ -1173,9 +1162,12 @@ func resourceServerImport(ctx context.Context, d *schema.ResourceData, meta inte
 	}
 
 	if len(parts) > 3 {
```
adeatcu-ionos, line 1164: @cristiGuranIonos can you add some info about how the ID would look like with more than 3 parts? Above we check if `len(parts) < 2` and you can see why since we have this:
```
Expecting {datacenter UUID}/{server UUID}"
```
So we kinda know what to expect. Here in the case of `len(parts) > 3` is unclear, I suppose it is something like `dcUUID/serverUUID/nicUUID/firewallRuleUUID` but I'm not sure, please add a comment related to this.

adeatcu-ionos (changes requested):

ionoscloud/firewall_service.go
```diff
@@ -0,0 +1,307 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	ionoscloud "github.com/ionos-cloud/sdk-go/v6"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/slice"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/utils"
+	"log"
+	"reflect"
+	"strconv"
+)
+
+type FirewallService struct {
```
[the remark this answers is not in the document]
cristiGuranIonos: circular dependencies. we have to rewrite some parts to be able to do it. We left it to be done at a later date.

ionoscloud/firewall_service.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+		if firewall.Properties.IcmpCode != nil {
+			fw["icmp_code"] = strconv.Itoa(int(*firewall.Properties.IcmpCode))
+		}
+	}
+	return fw
+}
+
+// GetChangesInFirewallRuleProperties - receives old new values from schema as interface and returns firewall properties
+// the schema testAccDataSourceContainerRegistryTokenMatchNameAndLocation contain duplicates between the old and new values
```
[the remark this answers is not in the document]
cristiGuranIonos: autocomplete did this. will change

ionoscloud/resource_server.go
```diff
@@ -1173,9 +1162,12 @@ func resourceServerImport(ctx context.Context, d *schema.ResourceData, meta inte
...
-		if err := d.Set("firewallrule_id", parts[3]); err != nil {
-			return nil, fmt.Errorf("error setting firewallrule_id %w", err)
+		var rules []string
+		rules = append(rules, parts[3])
+		if err = setFirewallRulesInSchema(d, rules); err != nil {
```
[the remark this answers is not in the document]
cristiGuranIonos: we currently support only one nic and one firewall rule id imported. We actually need to rethink this once we also have multiple nics. But for now, we would stick with one nicid and one fw id.

ionoscloud/resource_server.go
cristiGuranIonos: yes, that is how we support the import. will add a comment

ionoscloud/resource_server.go
```diff
@@ -603,73 +617,82 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, meta inte
...
-	firstNicProps := firstNicItem.Properties
-	if firstNicProps != nil {
-		firstNicIps := firstNicProps.Ips
-		if firstNicIps != nil &&
-			len(*firstNicIps) > 0 {
-			log.Printf("[DEBUG] set primary_ip to %s", (*firstNicIps)[0])
-			if err := d.Set("primary_ip", (*firstNicIps)[0]); err != nil {
-				diags := diag.FromErr(fmt.Errorf("error while setting primary ip %s: %w", d.Id(), err))
-				return diags
+				volumeItems := serverReq.Entities.Volumes.Items
```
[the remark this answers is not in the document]
cristiGuranIonos: We should be using orderedRuleIds here to keep the order.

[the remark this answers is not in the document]
cristiGuranIonos: also yes, we should move to serverReq when setting properties

ionoscloud/resource_server.go
```diff
@@ -1427,58 +1423,54 @@ func setResourceServerData(ctx context.Context, client *ionoscloud.APIClient, d
 	}
 
 	_, primaryNicOk := d.GetOk("primary_nic")
```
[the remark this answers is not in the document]
cristiGuranIonos: so actually here the problem might be the subsequent calls if we don't have a  primary_nic. Will test

ionoscloud/firewall_service.go
```diff
@@ -0,0 +1,307 @@
...
+func GetChangesInFirewallRuleProperties(oldValues, newValues []interface{}) ([]ionoscloud.FirewallruleProperties, []ionoscloud.FirewallruleProperties, error) {
```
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos I see that the description was changed and now it's clearer, but I would still change the function name to make it even more clearer but do as you feel.

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos not sure what you mean by your answers. More than that, I see that my comment doesn't make sense in the first place, we can't set the info about volumes using `foundFirstNic` because the `nic` doesn't contain information about volumes, I will mark this as solved for now since my original comment didn't make sense.

ionoscloud/resource_server.go
adeatcu-ionos: @cristiGuranIonos the comment it's not added yet.

ionoscloud/resource_server.go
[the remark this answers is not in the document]
adeatcu-ionos: @cristiGuranIonos ok, but I would still add a comment about this in the code.

ionoscloud/resource_server.go
```diff
@@ -923,130 +919,127 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
-		for _, n := range *server.Entities.Nics.Items {
-			nicStr := d.Get("primary_nic").(string)
-			if *n.Id == nicStr {
-				nic = &n
-				break
+		_, nicIntf := d.GetChange("nic")
+		newNics := nicIntf.([]interface{})
+		if len(newNics) == 0 {
+			err = d.Set("nic", nil)
+			if err := d.Set("primary_nic", ""); err != nil {
```
adeatcu-ionos, line 926: @cristiGuranIonos this check is duplicated, please remove one of them.

ionoscloud/resource_server.go
```diff
@@ -1423,56 +1427,56 @@ func setResourceServerData(ctx context.Context, client *ionoscloud.APIClient, d
...
-
 	if len(firewallRuleIds) == 0 {
 		if err := d.Set("firewallrule_id", ""); err != nil {
 			return err
 		}
-		if err := d.Set("firewallrule_ids", firewallRuleIds); err != nil {
-			return err
-		}
 	}
-	// Labels logic
```
adeatcu-ionos, line 1496: @cristiGuranIonos I don't think you wanted to remove this so please add it back.

adeatcu-ionos (changes requested):

ionoscloud/resource_server.go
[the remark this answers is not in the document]
cristiGuranIonos: added for the function:
// resourceServerImport can be either ionoscloud_server.myserver {datacenter uuid}/{server uuid} or  ionoscloud_server.myserver {datacenter uuid}/{server uuid}/{primary nic id}/{firewall rule id}

ionoscloud/resource_server.go
```diff
@@ -128,22 +137,23 @@ func resourceServer() *schema.Resource {
...
-					sshKeys := d.Get("volume.0.ssh_keys").([]interface{})
-					oldSshKeys := d.Get("ssh_keys").([]interface{})
-
-					if len(utils.DiffSlice(convertSlice(sshKeys), convertSlice(oldSshKeys))) == 0 {
-						return true
-					}
-
-					return false
-				},
+				//todo: remove as test servervassic fails
```
adeatcu-ionos, line 140: @cristiGuranIonos not sure if this needs to be removed or not, check it out please.

ionoscloud/resource_server.go
adeatcu-ionos: @cristiGuranIonos this is not solved yet.

ionoscloud/resource_server.go
```diff
@@ -1001,15 +1015,15 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 			mProp, _ := json.Marshal(nicProperties)
 			log.Printf("[DEBUG] Updating props: %s", string(mProp))
 			var createdNic ionoscloud.Nic
-			if !updateNic {
+			if createNic {
 				nic.Properties = &nicProperties
 				createdNic, apiResponse, err = client.NetworkInterfacesApi.DatacentersServersNicsPost(ctx, d.Get("datacenter_id").(string), *server.Id).Nic(*nic).Execute()
-			} else {
+			} else if nic.Id != nil {
```
adeatcu-ionos, line 1021: @cristiGuranIonos is this check really needed? not sure if there will be a case in which we want to update a `nic` and the `nic` has no ID.

ionoscloud/resource_server.go
```diff
@@ -1001,15 +1015,15 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
 				_, apiResponse, err = client.NetworkInterfacesApi.DatacentersServersNicsPatch(ctx, d.Get("datacenter_id").(string), *server.Id, *nic.Id).Nic(nicProperties).Execute()
 			}
 			logApiRequestTime(apiResponse)
 			if err != nil {
-				diags := diag.FromErr(fmt.Errorf("error updating nic (%s)", err))
+				diags := diag.FromErr(fmt.Errorf("error nic (%w)", err))
```
adeatcu-ionos, line 1026: @cristiGuranIonos why did we remove `updating` from here? `error updating nic: %w` makes sense to me

ionoscloud/resource_server.go
```diff
@@ -1032,14 +1046,14 @@ func resourceServerUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 				return diag.FromErr(err)
 			}
 
-			_, primaryNicOk := d.GetOk("primary_nic")
-			if createdNic.Id != nil && !updateNic && !primaryNicOk {
+			if createdNic.Id != nil && createNic {
```
adeatcu-ionos, line 1049: @cristiGuranIonos not sure here, but I assume that if `createNic` is true, then `createdNic` will be always different than `nil`, so not sure if the first check related to the ID makes sense.

adeatcu-ionos (changes requested):

ionoscloud/resource_server.go
cristiGuranIonos: This comment doesn't add much in terms of info since it's clear it's label logic below it, but i'll add it back

ionoscloud/resource_server.go
cristiGuranIonos: will remain to be investigated. Found a separate problem related to this diff.

ionoscloud/resource_server.go
cristiGuranIonos: it's not clear if it's an update or a create that errored.

ionoscloud/resource_server.go
cristiGuranIonos: just to be sure, but if createNic is set wrong in some corner case, I don't want it to crash

ionoscloud/resource_server.go
cristiGuranIonos: to make sure it doesn't crash and it throws a normal error.

adeatcu-ionos (approved):

PR #450, opened 2023-08-28
fix: increase notfoundchecks to 9999 for k8s large clusters

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/services/cloudapi/state.go
+++ b/services/cloudapi/state.go
@@ -3,11 +3,12 @@ package cloudapi
 import (
 	"context"
 	"fmt"
+	"log"
+	"time"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services"
-	"log"
-	"time"
 )
 
 // GetStateChangeConf gets the default configuration for tracking a request progress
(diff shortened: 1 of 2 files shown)


services/cloudapi/state.go
```diff
@@ -18,8 +19,8 @@ func GetStateChangeConf(meta interface{}, d *schema.ResourceData, location strin
 		Refresh:        resourceStateRefreshFunc(meta, location),
 		Timeout:        d.Timeout(timeoutType),
 		MinTimeout:     5 * time.Second,
-		Delay:          0,   // Don't delay the start
-		NotFoundChecks: 600, //Setting high number, to support long timeouts
+		Delay:          0,    // Don't delay the start
+		NotFoundChecks: 9999, //Setting high number, to support long timeouts
```
adeatcu-ionos, line 23: @cristiGuranIonos the title contains `for k8s large clusters` but it seems like the change affects all components that use cloud API service, is this intended? shouldn't we do this only for k8s clusters?

adeatcu-ionos (changes requested):

services/cloudapi/state.go
cristiGuranIonos: This only affects when setting a really high timeout. So even if practically it will only get used for k8s(because nothing else takes this long to create), it makes sense to change it for all cloudapi resources.

adeatcu-ionos (dismissed):

services/cloudapi/state.go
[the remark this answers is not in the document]
cristiGuranIonos: basically this will fail on the context timeout for create/update/delete etc. So it doesn't really matter if we choose a smaller number here.

PR #477, opened 2023-10-19
feat: add flowlog to nic

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_nic.go
+++ b/ionoscloud/resource_nic.go
@@ -133,7 +133,7 @@ func ForceNewForFlowlogChanges(_ context.Context, d *schema.ResourceDiff, _ inte
 		oldFLowLogs := oldVal.([]any)
 		newFlowLogs := newVal.([]any)
 
-		// new flowlog added during an upgrade
+		// "this check is for the scenario in which we have 0 initial flowlogs and we add a new one during a nic update
 		if (oldFLowLogs == nil || len(oldFLowLogs) == 0) && (newFlowLogs != nil || len(newFlowLogs) > 0) {
 			return nil
 		}
--- a/ionoscloud/resource_nic.go
+++ b/ionoscloud/resource_nic.go
@@ -137,6 +137,7 @@ func ForceNewForFlowlogChanges(_ context.Context, d *schema.ResourceDiff, _ inte
 		if (oldFLowLogs == nil || len(oldFLowLogs) == 0) && (newFlowLogs != nil || len(newFlowLogs) > 0) {
 			return nil
 		}
+		// flowlog deleted from resource
 		if (oldFLowLogs != nil || len(oldFLowLogs) > 0) && (newFlowLogs == nil || len(newFlowLogs) == 0) {
 			return d.ForceNew("flowlog")
 		}
@@ -148,9 +149,9 @@ func ForceNewForFlowlogChanges(_ context.Context, d *schema.ResourceDiff, _ inte
 		if newFlowLogs != nil && len(newFlowLogs) > 0 {
 			newFlowLogMap = newFlowLogs[0].(map[string]any)
 		}
-		// find the diff between the old and new value.
+		// find the diff between the old and new value of the fields.
 		// name should not force re-creation
-		// all other values should force re-creation, but case does not matter
+		// all other values should force re-creation, case does not matter
 		for k, v := range newFlowLogMap {
 			if k != "name" && k != "id" {
 				if !strings.EqualFold(strings.ToUpper(v.(string)), strings.ToUpper(oldFlowlogMap[k].(string))) {
--- a/ionoscloud/resource_nic.go
+++ b/ionoscloud/resource_nic.go
@@ -107,10 +107,11 @@ func resourceNic() *schema.Resource {
 				Optional: true,
 				Elem:     cloudapinic.FlowlogSchemaResource,
 				MaxItems: 1,
-				Description: `Flow logs holistically capture network information such as source and destination 
-							IP addresses, source and destination ports, number of packets, amount of bytes, 
-							the start and end time of the recording, and the type of protocol – 
-							and log the extent to which your instances are being accessed.`,
+				Description: `Only 1 flow log can be configured. After configuration, the flowlog's fields cannot be updated without re-creating the resource. 
+Flow logs holistically capture network information such as source and destination 
+IP addresses, source and destination ports, number of packets, amount of bytes, 
+the start and end time of the recording, and the type of protocol – 
+and log the extent to which your instances are being accessed.`,
 			},
 		},
 		Timeouts: &resourceDefaultTimeouts,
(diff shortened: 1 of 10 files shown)


utils/utils.go
```diff
@@ -322,3 +322,36 @@ func IsValueInSliceOfMap[T comparable](sliceOfMaps []interface{}, key string, va
...
+	if err != nil {
+		return result, err
+	}
+	// convert camelcase to snake-case
+	for k, v := range result {
+		if !strings.EqualFold(k, strcase.ToSnake(k)) {
+			result[strcase.ToSnake(k)] = v
+			delete(result, k)
+		}
+	}
```
avirtopeanu-ionos, line 350: I think `Decode` should be called before converting to snake-case

utils/utils.go
```diff
@@ -322,3 +322,36 @@ func IsValueInSliceOfMap[T comparable](sliceOfMaps []interface{}, key string, va
...
+
+	return result, decoder.Decode(input)
```
avirtopeanu-ionos, line 352:
```suggestion
	err = decoder.Decode(input)
	if err != nil {
		return nil, err
	}

	newResult := make(map[string]interface{})
	for k, v := range result {
		newResult[strcase.ToSnake(k)] = v
	}

	return newResult, nil
```

ionoscloud/resource_nic.go
```diff
@@ -102,6 +102,17 @@ func resourceNic() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"flowlog": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     cloudapinic.FlowlogSchemaResource,
+				MaxItems: 1,
+				Description: `Only 1 flow log can be configured. After configuration, the flowlog's fields cannot be updated without re-creating the resource.
```
adeatcu-ionos, line 110: @cristiGuranIonos fix indentation here please.

ionoscloud/resource_nic_test.go
```diff
@@ -231,6 +237,12 @@ resource ` + constant.NicResource + ` "database_nic" {
   firewall_type = "INGRESS"
   ips = [ ionoscloud_ipblock.test_server.ips[0], ionoscloud_ipblock.test_server.ips[1] ]
   name = "%s"
+  flowlog {
+    name = "test_flowlog"
```
adeatcu-ionos, line 241: @cristiGuranIonos not sure if this is required or not but I saw that we have multiple options that can be configured here, maybe we also want to have multiple tests for those options but I am not sure.

services/cloudapi/cloudapinic/nic.go
```diff
@@ -6,11 +6,77 @@ import (
...
+			Description: "Specifies the traffic direction pattern. Valid values: INGRESS, EGRESS, BIDIRECTIONAL. Immutable, forces re-recreation of the nic resource.",
+			Required:    true,
+			ForceNew:    true,
+		},
+		"name": {
+			Type:             schema.TypeString,
+			Description:      "The resource name. Immutable, forces re-recreation of the nic resource.",
+			Required:         true,
+			ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
+			ForceNew:         true,
```
adeatcu-ionos, line 45: @cristiGuranIonos I'm just double checking that we intend to have `ForceNew` here for the `name` attribute since this is not happening for other resources.

services/cloudapi/cloudapinic/nic.go
```diff
@@ -267,5 +355,21 @@ func NicSetData(d *schema.ResourceData, nic *ionoscloud.Nic) error {
 		}
 	}
 
+	if nic.Entities != nil && nic.Entities.Flowlogs != nil && nic.Entities.Flowlogs.Items != nil && len(*nic.Entities.Flowlogs.Items) > 0 {
```
adeatcu-ionos, line 358: @cristiGuranIonos shouldn't we set the `flowlog` to an empty set when `len(*nic.Entities.Flowlogs.Items) == 0`? Right now it seems that we don't set anything if the set is empty.

services/cloudapi/cloudapinic/nic.go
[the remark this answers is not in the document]
cristiGuranIonos: nothing can be updated on the flowlog from the documentation.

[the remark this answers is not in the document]
cristiGuranIonos: after more tests, seem that the name can be updated without replacement

ionoscloud/resource_nic_test.go
adeatcu-ionos: solving this since the whole resource has been rewritten

ionoscloud/resource_nic_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build compute || all || nic
-
```
adeatcu-ionos, line 2: @cristiGuranIonos shouldn't we add cases for all the scenarios that can occur related to changes/modifications? For example when you have a flowlog and you delete it, or when you modify a `force-new` field and so on.

services/cloudapi/cloudapinic/nic.go
cristiGuranIonos: the entire nic will be deleted when the flowlog is empty, so it's not necessary

ionoscloud/resource_nic_test.go
cristiGuranIonos: We have an update part below:
  flowlog {
    name = "test_flowlog_updated"
    action = "REJECTED"
    direction = "EGRESS"
    bucket = "` + constant.FlowlogBucketUpdated + `"
  }
This part takes care of update(deleting the nic also). We also have the last step of the test where the flowlog is missing which again forces re-creation of the nic and checks that it no longer exists.

ionoscloud/resource_nic.go
```diff
@@ -102,11 +103,70 @@ func resourceNic() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"flowlog": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     cloudapiflowlog.FlowlogSchemaResource,
+				MaxItems: 1,
+				Description: `Only 1 flow log can be configured. Only the name field can change as part of an update. Flow logs holistically capture network information such as source and destination
```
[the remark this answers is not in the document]
cristiGuranIonos: multiline string, so we need the indentation or it looks weird when it is shown to the user

ionoscloud/resource_nic_test.go
cristiGuranIonos: testAccCheckNicConfigUpdate - takes care of flowlog update
testAccCheckNicConfigUpdateIpv6 - takes care of deletion

adeatcu-ionos (approved):

PR #483, opened 2023-11-02
feat: add flowlog to alb

cristiGuranIonos:
## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/resources/application_loadbalancer.md
+++ b/docs/resources/application_loadbalancer.md
@@ -16,31 +16,30 @@ Manages an **Application Load Balancer** on IonosCloud.
 ```hcl
 
 resource "ionoscloud_datacenter" "example" {
-    name                  = "Datacenter Example"
-    location              = "us/las"
-    description           = "datacenter description"
-    sec_auth_protection   = false
+  name                  = "Datacenter Example"
+  location              = "us/las"
+  description           = "datacenter description"
+  sec_auth_protection   = false
 }
 
 resource "ionoscloud_lan" "example_1" {
-    datacenter_id         = ionoscloud_datacenter.example.id
-    public                = true
-    name                  = "Lan Example"
+  datacenter_id = ionoscloud_datacenter.example.id
+  public        = true
+  name          = "Lan Example"
 }
 
 resource "ionoscloud_lan" "example_2" {
-    datacenter_id         = ionoscloud_datacenter.example.id
-    public                = true
-    name                  = "Lan Example"
+  datacenter_id = ionoscloud_datacenter.example.id
+  public        = true
+  name          = "Lan Example"
 }
 
 resource "ionoscloud_application_loadbalancer" "example" {
-    datacenter_id         = ionoscloud_datacenter.example.id
-    name                  = "ALB Example"
-    listener_lan          = ionoscloud_lan.example_1.id
-    ips                   = [ "10.12.118.224"]
-    target_lan            = ionoscloud_lan.example_2.id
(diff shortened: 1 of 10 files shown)


docs/resources/application_loadbalancer.md
```diff
@@ -16,31 +16,30 @@ Manages an **Application Load Balancer** on IonosCloud.
...
-    datacenter_id         = ionoscloud_datacenter.example.id
-    name                  = "ALB Example"
-    listener_lan          = ionoscloud_lan.example_1.id
-    ips                   = [ "10.12.118.224"]
-    target_lan            = ionoscloud_lan.example_2.id
-    lb_private_ips        = [ "10.13.72.225/24"]
+  datacenter_id         = ionoscloud_datacenter.example.id
+  name                  = "ALB Example"
+  listener_lan          = ionoscloud_lan.example_1.id
+  ips                   = [ "10.12.118.224"]target_lan            = ionoscloud_lan.example_2.id
```
cbeti-ionos, line 41: newline before `target_lan`

ionoscloud/resource_application_loadbalancer.go
```diff
@@ -227,6 +267,34 @@ func resourceApplicationLoadBalancerUpdate(ctx context.Context, d *schema.Resour
 		request.Properties.LbPrivateIps = &privateIps
 	}
 
+	if d.HasChange("flowlog") {
+		old, newV := d.GetChange("flowlog")
```
cbeti-ionos, line 271: could use `oldV` to keep naming of old-new variable pairs consistent

ionoscloud/resource_application_loadbalancer.go
```diff
@@ -147,6 +159,25 @@ func resourceApplicationLoadBalancerCreate(ctx context.Context, d *schema.Resour
...
+			D:      d,
+			Client: client,
+		}
+		if flowLogList, ok := flowLogs.([]any); ok {
+			for _, flowLogData := range flowLogList {
+				if flowLogMap, ok := flowLogData.(map[string]interface{}); ok {
+					flowLog := cloudapiflowlog.GetFlowlogFromMap(flowLogMap)
+					err := fw.CreateOrPatchForALB(ctx, dcId, d.Id(), "", flowLog)
+					if err != nil {
+						d.SetId("")
```
cbeti-ionos, line 173: Should use d.Set("flowlog",nil) ?
Otherwise ALB will be deleted from state file even though it has been successfully created previously.

cristiGuranIonos: good catch!

cristiGuranIonos: there is a discussion if we shouldn't fail the entire resource though if flowlog creation fails when we create the entire resource.

cbeti-ionos: It would make sense to fail the entire load balancer creation, since you might want to capture all flow logs from the moment the alb resource is created.
Having only the flowlog creation fail, would it mean that if you add it to the lb later on in an update, you would miss all captures prior to that point?

cristiGuranIonos: That is correct. Failing the entire resource would still mean that you get the alb created in the backend unless we implement rollback.

cristiGuranIonos: implemented fallback

cbeti-ionos (dismissed):

PR #504, opened 2023-12-07
Feat: proxy protocol for alb and nlb

cbeti-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/resources/target_group.md
+++ b/docs/resources/target_group.md
@@ -22,6 +22,7 @@ resource "ionoscloud_target_group" "example" {
         ip                    = "22.231.2.2"
         port                  = "8080"
         weight                = "1"
+        proxy_protocol        = "v2ssl"
         health_check_enabled  = true
         maintenance_enabled   = false
     }    
@@ -29,6 +30,7 @@ resource "ionoscloud_target_group" "example" {
         ip                    = "22.231.2.3"
         port                  = "8081"
         weight                = "124"
+        proxy_protocol        = "v2"
         health_check_enabled  = false
         maintenance_enabled   = false
     }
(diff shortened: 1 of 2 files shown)


docs/resources/target_group.md
```diff
@@ -59,6 +59,7 @@ The following arguments are supported:
     - `ip` - (Required)[string] The IP of the balanced target VM.
     - `port` - (Required)[int] The port of the balanced target service; valid range is 1 to 65535.
     - `weight` - (Required)[int] Traffic is distributed in proportion to target weight, relative to the combined weight of all targets. A target with higher weight receives a greater share of traffic. Valid range is 0 to 256 and default is 1; targets with weight of 0 do not participate in load balancing but still accept persistent connections. It is best use values in the middle of the range to leave room for later adjustments.
+    - `proxy_protocol` - (Optional)[string] The proxy protocol version. Accepted values are `none`, `v1`, `v2`, `v2ssl`. If unspecified, the default value of `none` is used.
```
adeatcu-ionos, line 62: @cbeti-ionos I suggest also adding this in the example as you did for `docs/resources/networkloadbalancer_forwardingrule.md`

ionoscloud/resource_target_group_test.go
```diff
@@ -50,6 +51,7 @@ func TestAccTargetGroupBasic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceNameTargetGroup, "targets.0.ip", "22.231.2.2"),
 					resource.TestCheckResourceAttr(resourceNameTargetGroup, "targets.0.port", "8080"),
 					resource.TestCheckResourceAttr(resourceNameTargetGroup, "targets.0.weight", "1"),
+					resource.TestCheckResourceAttr(resourceNameTargetGroup, "targets.0.proxy_protocol", "v2"),
```
adeatcu-ionos, line 54: @cbeti-ionos it doesn't have anything to do with the scope of this PR, but since we are here I would also suggest adding multiple `targets` for these tests, it's good to have multiple `targets` since in the past involving multiple resources exposed various bugs. Right now, for `targets` we have only one element.

adeatcu-ionos (changes requested):

docs/resources/target_group.md
cbeti-ionos: done

ionoscloud/resource_target_group_test.go
cbeti-ionos: added another target and also tested it's correct removal at the end

adeatcu-ionos (approved):

PR #520, opened 2024-01-19
feat: Add support for MariaDB, add tests and docs

adeatcu-ionos:
## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/data_source_dbaas_mariadb_cluster.go
+++ b/ionoscloud/data_source_dbaas_mariadb_cluster.go
@@ -1,0 +1,158 @@
+package ionoscloud
+
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	dbaas "github.com/ionos-cloud/sdk-go-dbaas-maria"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services"
+	"strings"
+)
+
+func dataSourceDBaaSMariaDBCluster() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMariaDBClusterRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:             schema.TypeString,
+				Description:      "The id of your cluster.",
+				Optional:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.IsUUID),
+			},
+			"display_name": {
+				Type:        schema.TypeString,
+				Description: "The friendly name of your cluster.",
+				Optional:    true,
+			},
+			"mariadb_version": {
+				Type:        schema.TypeString,
+				Description: "The MariaDB version of your cluster.",
+				Computed:    true,
+			},
+			"instances": {
+				Type:        schema.TypeInt,
+				Description: "The total number of instances in the cluster (one primary and n-1 secondary).",
+				Computed:    true,
+			},
+			"cores": {
+				Type:        schema.TypeInt,
+				Description: "The number of CPU cores per instance.",
+				Computed:    true,
+			},
+			"ram": {
+				Type:        schema.TypeInt,
+				Description: "The amount of memory per instance in gigabytes (GB).",
+				Computed:    true,
+			},
+			"storage_size": {
+				Type:        schema.TypeInt,
+				Description: "The amount of storage per instance in gigabytes (GB).",
+				Computed:    true,
+			},
+
+			"connections": {
+				Type:        schema.TypeList,
+				Description: "The network connection for your cluster. Only one connection is allowed.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"datacenter_id": {
+							Type:        schema.TypeString,
+							Description: "The datacenter to connect your cluster to.",
+							Computed:    true,
+						},
+						"lan_id": {
+							Type:        schema.TypeString,
+							Description: "The numeric LAN ID to connect your cluster to.",
+							Computed:    true,
+						},
+						"cidr": {
+							Type:        schema.TypeString,
+							Description: "The IP and subnet for your cluster.",
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"maintenance_window": {
+				Type:        schema.TypeList,
+				Description: "A weekly 4 hour-long window, during which maintenance might occur.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"time": {
+							Type:        schema.TypeString,
+							Description: "Start of the maintenance window in UTC time.",
+							Computed:    true,
+						},
+						"day_of_the_week": {
+							Type:        schema.TypeString,
+							Description: "The name of the week day.",
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"dns_name": {
+				Type:        schema.TypeString,
+				Description: "The DNS name pointing to your cluster.",
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func dataSourceMariaDBClusterRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(services.SdkBundle).MariaDBClient
+	id, idOk := d.GetOk("id")
+	name, nameOk := d.GetOk("display_name")
+
+	if idOk && nameOk {
+		return diag.FromErr(fmt.Errorf("ID and display_name cannot be both specified at the same time"))
+	}
+	if !idOk && !nameOk {
+		return diag.FromErr(fmt.Errorf("please provide either the MariaDB cluster ID or display_name"))
+	}
+
+	var cluster dbaas.ClusterResponse
+	var err error
+
+	if idOk {
+		/* search by ID */
+		cluster, _, err = client.GetCluster(ctx, id.(string))
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("an error occurred while fetching the MariaDB cluster with ID %v: %w", id.(string), err))
+		}
+	} else {
+		clusters, _, err := client.ListClusters(ctx, "")
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("an error occurred while fetching MariaDB clusters: %w", err))
+		}
+
+		var results []dbaas.ClusterResponse
+
+		if clusters.Items != nil && len(*clusters.Items) > 0 {
+			for _, clusterItem := range *clusters.Items {
+				if clusterItem.Properties != nil && clusterItem.Properties.DisplayName != nil && strings.EqualFold(*clusterItem.Properties.DisplayName, name.(string)) {
+					results = append(results, clusterItem)
+				}
+			}
+		}
+
+		if results == nil || len(results) == 0 {
+			return diag.FromErr(fmt.Errorf("no MariaDB cluster found with the specified display name: %v", name))
+		} else if len(results) > 1 {
+			return diag.FromErr(fmt.Errorf("more than one MariaDB cluster found with the specified criteria name: %v", name))
+		} else {
+			cluster = results[0]
+		}
+
+	}
+
+	if err := client.SetMariaDBClusterData(d, cluster); err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
+}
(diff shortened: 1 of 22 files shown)


adeatcu-ionos:
**NOTE:** This code is not tested yet since the API is not working. After fixing the API, we will come back to this and test it.

ionoscloud/data_source_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	} else {
+		clusters, _, err := client.ListClusters(ctx, "")
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("an error occurred while fetching MariaDB clusters: %w", err))
+		}
+
+		var results []dbaas.ClusterResponse
+
+		if clusters.Items != nil && len(*clusters.Items) > 0 {
```
cbeti-ionos, line 138: can shorten to just `if clusters.Items != nil` since there is only the for loop under this condition

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,245 @@
...
+			},
+		},
+		Timeouts: &resourceDefaultTimeouts,
+	}
+}
+
+func mariaDBClusterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(services.SdkBundle).MariaDBClient
+
+	cluster, err := dbaasService.GetMariaDBClusterDataCreate(d)
```
cbeti-ionos, line 156: naming seems a bit off: both 'Get' and 'Create' 
also, we are using both a `dbaasService` and a `MariaDBClient` objects to perform db functionality so maybe we should integrate these into the same service

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,245 @@
...
+	d.SetId(clusterID)
+
+	err = utils.WaitForResourceToBeReady(ctx, d, client.IsClusterReady)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error occured while checking the status for MariaDB cluster with ID: %v, error: %w", clusterID, err))
+	}
+	if err := client.SetMariaDBClusterData(d, response); err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
```
cbeti-ionos, line 175: call `mariaDBClusterRead` on non-error return from CreateContext

ionoscloud/resource_dbaas_mariadb_cluster_test.go
```diff
@@ -0,0 +1,225 @@
...
+const clusterVersionValue = "10.6"
+const clusterInstancesValue = "1"
+const clusterCoresValue = "4"
+const clusterRamValue = "2"
+const clusterStorageSizeValue = "10"
+const clusterConnectionsCidrValue = "192.168.1.100/24"
+const clusterDisplayNameValue = constant.DBaaSClusterTestResource
+const clusterCredentialsUsernameValue = "username"
+const datacenterResourceName = "datacenter_example"
+const lanResourceName = "lan_example"
```
cbeti-ionos, line 208: we can group these in a const block, and maybe move them at the top or bottom
that way we group single-values separately from hcl blocks

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,245 @@
...
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services"
+	dbaasService "github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services/dbaas"
```
cbeti-ionos, line 13: `dbaasService` is not quite idiomatic for custom go import declarations
we should do some refactoring to import all we need only once from the `services` package

services/dbaas/cluster.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	mariadb "github.com/ionos-cloud/sdk-go-dbaas-maria"
```
cbeti-ionos, line 14: We should really start subpackaging these dbaas files on a per-product basis otherwise they will become unmaintainable as more database implementations are added.

ionoscloud/data_source_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,158 @@
...
+				Type:        schema.TypeString,
+				Description: "The MariaDB version of your cluster.",
+				Computed:    true,
+			},
+			"instances": {
+				Type:        schema.TypeInt,
+				Description: "The total number of instances in the cluster (one primary and n-1 secondary).",
+				Computed:    true,
+			},
+			"cores": {
```
cristiGuranIonos, line 39: Do we have any validation for this, ram and storage_size?

cristiGuranIonos: nevermind, thought this was the resource

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,238 @@
...
+	client := meta.(services.SdkBundle).MariaDBClient
+	clusterId := d.Id()
+	cluster, apiResponse, err := client.GetCluster(ctx, clusterId)
+
+	if err != nil {
+		if apiResponse.HttpNotFound() {
+			d.SetId("")
+			return nil, fmt.Errorf("MariaDB cluster with ID: %v does not exist", clusterId)
+		}
+		return nil, fmt.Errorf("an error occured while trying to import MariaDB cluster with ID: %v", clusterId)
```
cristiGuranIonos, line 204: I think we need to actually throw the error

services/dbaas/client.go
```diff
@@ -9,35 +9,40 @@ import (
...
-	newConfigDbaas := psql.NewConfiguration(username, password, token, url)
+type PsqlClient struct {
+	sdkClient *psql.APIClient
+}
 
-	if os.Getenv(constant.IonosDebug) != "" {
+func NewMariaDBClient(username, password, token, url, version, terraformVersion string) *MariaDBClient {
+	newConfigDbaas := mariadb.NewConfiguration(username, password, token, url)
+
+	if os.Getenv("IONOS_DEBUG") != "" {
```
cristiGuranIonos, line 33: IonosDebug can be used i think

services/dbaas/cluster.go
```diff
@@ -825,6 +990,102 @@ func SetPgSqlClusterData(d *schema.ResourceData, cluster psql.ClusterResponse) e
...
+	connectionMap := map[string]interface{}{}
+
+	utils.SetPropWithNilCheck(connectionMap, "datacenter_id", connection.DatacenterId)
+	utils.SetPropWithNilCheck(connectionMap, "lan_id", connection.LanId)
+	utils.SetPropWithNilCheck(connectionMap, "cidr", connection.Cidr)
+
+	return connectionMap
+}
+
+func (c *MariaDBClient) SetMaintenanceWindowProperties(maintenanceWindow mariadb.MaintenanceWindow) map[string]interface{} {
```
cristiGuranIonos, line 1080: we should be able to use an interface here based on the getters. For another time.

cristiGuranIonos (dismissed):

ionoscloud/data_source_dbaas_mariadb_cluster.go
adeatcu-ionos: @cbeti-ionos done in eabe0a8e1529e0c3b0c3d8b882cc4b2d6056674f

ionoscloud/resource_dbaas_mariadb_cluster.go
adeatcu-ionos: @cbeti-ionos I agree with this and also with the comments that you provided related to the structure of the packages. I will leave this as it is for now and as soon as this branch is merged, I will start working on separating the dbaas services into different packages and make all the refactor/renaming work. I decided this because splitting dbaas services into different packages it's not the main purpose of this PR.

ionoscloud/resource_dbaas_mariadb_cluster.go
adeatcu-ionos: @cbeti-ionos is there a reason behind this? We are already setting the data into the state using `SetMariaDBClusterData`, not sure how an extra call to `mariaDBClusterRead` would help here.

ionoscloud/resource_dbaas_mariadb_cluster_test.go
adeatcu-ionos: @cbeti-ionos done in eabe0a8e1529e0c3b0c3d8b882cc4b2d6056674f

ionoscloud/resource_dbaas_mariadb_cluster.go
cbeti-ionos: I think there might be a difference in the completeness of the data, since you use the response from a `client.CreateCluster` call and the cluster itself is not fully provisioned at this point.

services/dbaas/client.go
adeatcu-ionos: @adeatcu-ionos solved in a25d7c6c779850ffeb70f942e90a214806ee271d

ionoscloud/resource_dbaas_mariadb_cluster.go
adeatcu-ionos: @cristiGuranIonos solved in a25d7c6c779850ffeb70f942e90a214806ee271d

ionoscloud/resource_dbaas_mariadb_cluster.go
adeatcu-ionos: @cbeti-ionos `I think there might be a difference in the completeness of the data` hmm, it seems like a valid point, I will check if this is the case when the API will work, let's keep this open until then.

adeatcu-ionos:
@cbeti-ionos @cristiGuranIonos please take another look, I added another commit that contains functionality for backups, the tests for the backups data source are still missing, I will add them when the API will work.

ionoscloud/resource_dbaas_mariadb_cluster.go
cbeti-ionos: I would also go a little bit further and say it would be a better separation of concerns to not have the Create function also set state data (wherever and whenever possible, we still need to set things like ID there), and delegate that responsibility entirely to the Reader and simply call it when we want state writing.

adeatcu-ionos: @cbeti-ionos the API works now, there is no problem from the data completion POV, but let's discuss tomorrow about this.

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,238 @@
...
+		UpdateContext: mariaDBClusterUpdate,
+		Importer: &schema.ResourceImporter{
+			StateContext: mariaDBClusterImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"mariadb_version": {
+				Type:             schema.TypeString,
+				Description:      "The MariaDB version of your cluster.",
+				Required:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.All(validation.StringInSlice([]string{"10.6"}, true), validation.StringIsNotWhiteSpace)),
```
cristiGuranIonos, line 32: this validation would mean that if/when the version changes or a new one is added, we need to do a new release of tf to support it.

ionoscloud/resource_dbaas_mariadb_cluster.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		return diag.FromErr(fmt.Errorf("error while deleting MariaDB cluster with ID: %v, error: %w", clusterID, err))
+	}
+	err = utils.WaitForResourceToBeDeleted(ctx, d, client.IsClusterDeleted)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("deletion check failed for MariaDB cluster with ID: %v, error: %w", clusterID, err))
+	}
+
+	// wait 15 seconds after the deletion of the cluster, for the lan to be freed
+	time.Sleep(constant.SleepInterval * 10)
```
cristiGuranIonos, line 188: we should remove this. If this is absolutely necessary, we should add a todo comment and say that we need to revisit it in the future

cristiGuranIonos (dismissed):

PR #534, opened 2024-03-18
refactor: Add 'workflow_dispatch' option in workflows

adeatcu-ionos:

There are cases in which we want to run a specific pipeline after we merge a fix in master and we don't want to wait until next scheduled run, this is why I added the possibility to run the pipeline manually.

In this PR, I also removed the `schedule` option for `server` but this is only a temporary change.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


adeatcu-ionos:
I will close this PR since we can manually run whatever we want using `Test suite` but I completely forgot about it.

PR #541, opened 2024-04-09
feat: Add 'lans' property for Dataplatform clusters

adeatcu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


ionoscloud/data_source_dataplatform_cluster.go
```diff
@@ -269,16 +309,21 @@ func setAdditionalDataplatformClusterData(ctx context.Context, d *schema.Resourc
...
 	/* get from api and set in schema the kubeconfig*/
 	if cluster.Id != nil {
-		kubeConfig, _, err := client.GetClusterKubeConfig(ctx, *cluster.Id)
+		kubeConfigMap, _, err := client.GetClusterKubeConfig(ctx, *cluster.Id)
 		if err != nil {
 			return fmt.Errorf("an error occurred while fetching the kubernetes config for cluster with ID %s: %w", *cluster.Id, err)
 		}
 
-		if err := d.Set("kube_config", kubeConfig); err != nil {
+		kubeConfig, err := json.Marshal(kubeConfigMap)
```
adeatcu-ionos, line 317: _**NOTE:**_ The return type for `ClustersKubeconfigFindByClusterId` has changed and because of that `GetClusterKubeConfig` also returns a map instead of a string, this is why I'm making `json.Marshal` here.
_**BEFORE:**_

1. `ClustersKubeconfigFindByClusterId` -> `kube_config` string
2. Set `kube_config` schema attribute using the previously returned string
3. Unmarshal `kube_config` into a structure and populate `config` schema attribute using it.

_**AFTER:**_

1. `ClustersKubeconfigFindByClusterId` -> `kube_config` map
2. Marshal map into string
3. Set `kube_config` schema attribute with the previous string
5. Umarshal string into structure and populate `config` schema attribute using it

We should get rid of the `Unmarshal` method since we can operate on the returned map, but the code changes a little bit and this is not the scope of this PR so I did the `Marshal` in order to keep the current functionality.

services/dataplatform/versions.go
```diff
@@ -13,7 +13,7 @@ import (
 func (c *Client) GetVersions(ctx context.Context) ([]string, *dataplatform.APIResponse, error) {
 	versions, apiResponse, err := c.sdkClient.DataPlatformMetaDataApi.VersionsGet(ctx).Execute()
 	apiResponse.LogInfo()
-	return versions, apiResponse, err
+	return *versions.Items, apiResponse, err
```
cbeti-ionos, line 16: Did `VersionsGet` change its return type? Seems it initially returned []string directly but now it places it in the `Items` field?
We might need to check err before dereferencing now.

ionoscloud/data_source_dataplatform_cluster.go
```diff
@@ -269,16 +309,21 @@ func setAdditionalDataplatformClusterData(ctx context.Context, d *schema.Resourc
...
+		kubeConfigString := string(kubeConfig)
```
cristiGuranIonos, line 318: first check for error I think

services/dataplatform/cluster.go
```diff
@@ -129,9 +132,48 @@ func setPatchClusterRequestProperties(d *schema.ResourceData) *dataplatform.Patc
 		dataplatformCluster.Properties.MaintenanceWindow = setMaintenanceWindowData(d)
 	}
 
+	dataplatformCluster.Properties.Lans = setLansData(d)
+
 	return &dataplatformCluster
 }
 
+func setLansData(d *schema.ResourceData) *[]dataplatform.Lan {
```
cristiGuranIonos, line 140: do we need the pointer to the slice here? Can't we just return just the slice?

services/dataplatform/cluster.go
```diff
@@ -129,9 +132,48 @@ func setPatchClusterRequestProperties(d *schema.ResourceData) *dataplatform.Patc
...
+	var lansBody []dataplatform.Lan
+	if lansData, ok := d.GetOk("lans"); ok {
```
cristiGuranIonos, line 142: maybe we can check != ok here and below and just return early so we don't have so much nesting?

services/dataplatform/cluster.go
```diff
@@ -129,9 +132,48 @@ func setPatchClusterRequestProperties(d *schema.ResourceData) *dataplatform.Patc
...
+				}
+			}
+		}
+	}
+	return &lansBody
+}
+
+func setRoutesData(lan map[string]interface{}) *[]dataplatform.Route {
+	var routesBody []dataplatform.Route
+	if routesData, ok := lan["routes"].(*schema.Set); ok {
```
cristiGuranIonos, line 162: same

services/dataplatform/cluster.go
```diff
@@ -129,9 +132,48 @@ func setPatchClusterRequestProperties(d *schema.ResourceData) *dataplatform.Patc
...
+					lanBody.Routes = setRoutesData(lan)
+					lansBody = append(lansBody, lanBody)
+				}
+			}
+		}
+	}
+	return &lansBody
+}
+
+func setRoutesData(lan map[string]interface{}) *[]dataplatform.Route {
```
cristiGuranIonos, line 160: again, maybe just return slice

services/dataplatform/versions.go
adeatcu-ionos: @cbeti-ionos yes, before it was `[]string`, now it is `VersionsGet200Response`.

services/dataplatform/cluster.go
adeatcu-ionos: @cristiGuranIonos the code will look like this:
```
        var lansBody []dataplatform.Lan
	lansData, ok := d.GetOk("lans")
	if !ok {
		return lansBody
	}
	lansDataSet, ok := lansData.(*schema.Set)
	if !ok {
		return lansBody
	}
	for _, lanData := range lansDataSet.List() {
		if lan, ok := lanData.(map[string]interface{}); ok {
			lanID := lan["lan_id"].(string)
			dhcp := lan["dhcp"].(bool)
			var lanBody dataplatform.Lan
			lanBody.LanId = &lanID
			lanBody.Dhcp = &dhcp
			lanBody.Routes = setRoutesData(lan)
			lansBody = append(lansBody, lanBody)
		}
	}
	return lansBody
```
I prefer the first one with the nesting, I think it looks cleaner, what do you think?

services/dataplatform/cluster.go
adeatcu-ionos: @cristiGuranIonos same thing here, check the comment above

services/dataplatform/cluster.go
adeatcu-ionos: yes, we do:
```
dataplatformCluster.Properties.Lans = setLansData(d)
```
where `Lans` is a pointer to the slice:
```
Lans *[]Lan `json:"lans,omitempty"`
```
Same goes for all the similar comments, for `routes` for example.

services/dataplatform/versions.go
adeatcu-ionos: @cbeti-ionos done in cdd9b58070e612723abae0806410ff7d12151ea5

ionoscloud/data_source_dataplatform_cluster.go
adeatcu-ionos: @cristiGuranIonos done in cdd9b58070e612723abae0806410ff7d12151ea5

PR #550, opened 2024-04-24
test: Modify DBaaS workflow to run tests in multiple stages

adeatcu-ionos:

Since we have more DBaaS services now it's cleaner if we have different run stages for each service.
The timeout values were selected as follows:

1. `MariaDB` - Only one cluster is created, by default cluster creation timeout is 60 minutes so the test, in the worst case, shouldn't take more than 70 minutes.
2. `MongoDB` and `PgSQL` - Before we had a 240 minute timeout for both pipelines, so for each I thought about 120 minute but I decided to increase the value a little bit to make sure that there is plenty of time.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/.github/workflows/dbaas-test-run.yml
+++ b/.github/workflows/dbaas-test-run.yml
@@ -28,5 +28,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version-file: 'go.mod'
-      - name: Run tests with dbaas tag
-        run: go test ./ionoscloud -v -failfast -timeout 240m -tags dbaas
+      - name: Run tests with mariadb tag
+        run: go test ./ionoscloud -v -failfast -timeout 70m -tags mariadb
+      - name: Run tests with psql tag
+        run: go test ./ionoscloud -v -failfast -timeout 140m -tags psql
+      - name: Run tests with mongo tag
+        run: go test ./ionoscloud -v -failfast -timeout 140m -tags mongo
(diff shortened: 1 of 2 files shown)


.github/workflows/dbaas-test-run.yml
```diff
@@ -28,5 +28,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version-file: 'go.mod'
-      - name: Run tests with dbaas tag
-        run: go test ./ionoscloud -v -failfast -timeout 240m -tags dbaas
+      - name: Run tests with MariaDB tag
+        run: go test ./ionoscloud -v -failfast -timeout 70m -tags mariadb
+      - name: Run tests with PostgreSQL tag
```
cristiGuranIonos, line 33:
```suggestion
      - name: Run tests with psql tag
```

.github/workflows/dbaas-test-run.yml
```diff
@@ -28,5 +28,9 @@ jobs:
...
+        run: go test ./ionoscloud -v -failfast -timeout 140m -tags psql
+      - name: Run tests with MongoDB tag
```
cristiGuranIonos, line 35:
```suggestion
      - name: Run tests with mongo tag
```

cristiGuranIonos (approved):

PR #560, opened 2024-05-27
doc: Fix documentation issues

adeatcu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/resources/server.md
+++ b/docs/resources/server.md
@@ -22,6 +22,7 @@ data "ionoscloud_image" "example" {
     type                  = "HDD"
     cloud_init            = "V1"
     image_alias           = "ubuntu:latest"
+    location              = "us/las"
 }
 
 resource "ionoscloud_datacenter" "example" {
@@ -50,7 +51,7 @@ resource "ionoscloud_server" "example" {
     ram                   = 1024
     availability_zone     = "ZONE_1"
     cpu_family            = "AMD_OPTERON"
-    image_name            = data.ionoscloud_image.example.id
+    image_name            = data.ionoscloud_image.example.name
     image_password        = random_password.server_image_password.result
     type                  = "ENTERPRISE"
     volume {
--- a/docs/resources/volume.md
+++ b/docs/resources/volume.md
@@ -19,6 +19,7 @@ A primary volume will be created with the server. If there is a need for additio
 data "ionoscloud_image" "example" {
     type                  = "HDD"
     cloud_init            = "V1"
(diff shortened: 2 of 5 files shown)


docs/resources/server.md
```diff
@@ -22,6 +22,7 @@ data "ionoscloud_image" "example" {
     type                  = "HDD"
     cloud_init            = "V1"
     image_alias           = "ubuntu:latest"
+    location              = "us/ewr"
```
cristiGuranIonos, line 25: shouldn't this match the datacenter location? us/las?

docs/resources/volume.md
```diff
@@ -19,7 +19,8 @@ A primary volume will be created with the server. If there is a need for additio
 data "ionoscloud_image" "example" {
     type                  = "HDD"
     cloud_init            = "V1"
-    location              = "us/las"
+    image_alias           = "ubuntu:latest"
+    location              = "us/ewr"
```
cristiGuranIonos, line 23: same as above

docs/resources/server.md
adeatcu-ionos: @cristiGuranIonos no, it doesn't matter where the image is coming from.

docs/resources/volume.md
adeatcu-ionos: @cristiGuranIonos same as above, it doesn't matter when speaking about the image.

docs/resources/server.md
cristiGuranIonos: what if it's not available in the location?

cristiGuranIonos: I mean, if it's an example we should keep the location the same in the entire example, even if ubuntu:latest is available in all locations.

adeatcu-ionos: @cristiGuranIonos `what if it's not available in the location?` then the image will not be fetched, I don't think that we can reach a configuration to ensure that a image will always be available, whether we talk about `location`, `image_alias` or `type`. 

`if it's an example we should keep the location the same in the entire example, even if ubuntu:latest is available in all locations.` ok, I will change the location

adeatcu-ionos: @cristiGuranIonos done

docs/resources/volume.md
adeatcu-ionos: @cristiGuranIonos done

cristiGuranIonos (approved):

PR #566, opened 2024-06-12
feat: redis replica datasource

avirtopeanu-ionos:
Add Redis Replicaset datasource

### DIFF
--- a/ionoscloud/data_source_dbaas_redisdb_replicaset.go
+++ b/ionoscloud/data_source_dbaas_redisdb_replicaset.go
@@ -1,0 +1,199 @@
+package ionoscloud
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	redisdb "github.com/ionos-cloud/sdk-go-dbaas-redis"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services"
+	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/utils/constant"
+)
+
+func dataSourceDBaaSRedisDBReplicaSet() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceDBaaSRedisDBReplicaSetRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Description: "The ID of the Redis Replicaset.",
+				Optional:    true,
+			},
+			"location": {
+				Type:             schema.TypeString,
+				Description:      "The cluster location",
+				Optional:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(constant.MariaDBClusterLocations, false)),
+			},
+			"display_name": {
+				Type:        schema.TypeString,
+				Description: "The display name of the Redis Replicaset.",
+				Optional:    true,
+			},
+			"connections": {
+				Type:        schema.TypeList,
+				Description: "The network connection for your Replicaset. Only one connection is allowed.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"cidr": {
+							Type:        schema.TypeString,
+							Description: "The IP and subnet for your Replicaset.",
+							Computed:    true,
+						},
+						"datacenter_id": {
+							Type:        schema.TypeString,
+							Description: "The datacenter to connect your Replicaset to.",
+							Computed:    true,
+						},
+						"lan_id": {
+							Type:        schema.TypeString,
+							Description: "The numeric LAN ID to connect your Replicaset to.",
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"credentials": {
+				Type:        schema.TypeList,
+				Description: "The credentials for your Replicaset.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"password": {
+							Type:        schema.TypeString,
+							Description: "The password for your Replicaset.",
+							Computed:    true,
+						},
+						"username": {
+							Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Description: "The ID of the Redis Replicaset.",
+				Optional:    true,
+			},
+			"location": {
+				Type:             schema.TypeString,
+				Description:      "The cluster location",
+				Optional:         true,
```
adeatcu-ionos, line 28: @avirtopeanu-ionos `location` should be required, on MariaDB it is optional because of the poor design in the beginning, but here it should be required.

ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(constant.MariaDBClusterLocations, false)),
```
adeatcu-ionos, line 29: @avirtopeanu-ionos add a `TODO` here because we need to change the name of this constant.

ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+					},
+				},
+			},
+			"credentials": {
+				Type:        schema.TypeList,
+				Description: "The credentials for your Replicaset.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"password": {
```
adeatcu-ionos, line 71: @avirtopeanu-ionos not sure if this should be included here since we don't read anything about the password, add this `TODO` please: `TODO -- Check if 'password' needs to be here or not`

ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+	if !idOk && !nameOk {
+		return diag.FromErr(fmt.Errorf("please provide either the Redis replicaset ID or display_name"))
+	}
+
+	var cluster redisdb.ReplicaSetRead
+	var err error
+
+	if idOk {
+		// search by ID
+		cluster, _, err = client.GetReplicaSet(ctx, id.(string), location)
```
adeatcu-ionos, line 169: @avirtopeanu-ionos here it should be `replicaSet`, no `cluster`, please make this change wherever is needed.

ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+			},
+		},
+	}
+}
+
+func dataSourceDBaaSRedisDBReplicaSetRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(services.SdkBundle).RedisDBClient
+	id, idOk := d.GetOk("id")
+	name, nameOk := d.GetOk("display_name")
```
adeatcu-ionos, line 154: @avirtopeanu-ionos please change `name` to `displayName`, even though it's the same thing, let's name the var as the attribute.

ionoscloud/data_source_dbaas_redisdb_replicaset.go
```diff
@@ -0,0 +1,208 @@
...
+	location := d.Get("location").(string)
+
+	if idOk && nameOk {
+		return diag.FromErr(fmt.Errorf("ID and display_name cannot be both specified at the same time"))
+	}
+	if !idOk && !nameOk {
+		return diag.FromErr(fmt.Errorf("please provide either the Redis replicaset ID or display_name"))
+	}
+
+	var cluster redisdb.ReplicaSetRead
```
adeatcu-ionos, line 164: @avirtopeanu-ionos `replicaSet` instead of `cluster`.

services/dbaas/redisdb/replica_set.go
```diff
@@ -155,7 +156,13 @@ func (c *RedisDBClient) SetRedisDBReplicaSetData(d *schema.ResourceData, replica
 		}
 	}
 
-	// TODO -- Check if we need to set anything for 'credentials'
+	if replicaSet.Properties.Credentials != nil {
```
adeatcu-ionos, line 159: @avirtopeanu-ionos for `replicaSet.Properties.MaintenanceWindow` we have a different approach that looks cleaner, but you achieve the same thing using one line, so I think it's better to modify all the sets to use this format: `[]interface{}{setFunction(*replicaSet.Properties.Field)}`, please add a `TODO` for this.

adeatcu-ionos (changes requested):

services/dbaas/redisdb/replica_set.go
avirtopeanu-ionos: can we do this in the main branch?

ionoscloud/data_source_dbaas_redisdb_replicaset.go
avirtopeanu-ionos: since tihs is a nitpick, can we do this in the main branch?

ionoscloud/data_source_dbaas_redisdb_replicaset.go
avirtopeanu-ionos: you're correct, no need to add TODO, I'll remove this attribute

ionoscloud/data_source_dbaas_redisdb_replicaset.go
avirtopeanu-ionos: since this is a nitpick, can we do this in the main branch?

ionoscloud/data_source_dbaas_redisdb_replicaset.go
avirtopeanu-ionos: thanks, I'll add this

services/dbaas/redisdb/replica_set.go
adeatcu-ionos: @avirtopeanu-ionos all the `TODO`s are just some references for our PR. Before merging, no `TODO`s will be left.

PR #573, opened 2024-06-14
fix: save user group ids in state when reading or importing

digna-ionos:
- fetch group ids using the depth of 1
- store group ids in state when fetching user resources

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [x] Jira task updated

### DIFF
--- a/ionoscloud/resource_user.go
+++ b/ionoscloud/resource_user.go
@@ -162,11 +162,11 @@ func resourceUserCreate(ctx context.Context, d *schema.ResourceData, meta interf
 func resourceUserRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(services.SdkBundle).CloudApiClient
 
-	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, d.Id()).Execute()
+	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, d.Id()).Depth(1).Execute()
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
-		if apiResponse != nil && apiResponse.Response != nil && apiResponse.StatusCode == 404 {
+		if apiResponse.HttpNotFound() {
 			d.SetId("")
 			return nil
 		}
@@ -306,7 +306,7 @@ func resourceUserImporter(ctx context.Context, d *schema.ResourceData, meta inte
 
 	userId := d.Id()
 
-	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, userId).Execute()
+	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, userId).Depth(1).Execute()
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ionoscloud/resource_user_test.go
```diff
@@ -125,6 +125,16 @@ func TestAccUserBasic(t *testing.T) {
 						"name": "group1",
 					})),
 			},
+			{
+				Config:             testAccCheckNewUserGroup,
```
adeatcu-ionos, line 129: @digna-ionos it isn't clear for me what we are testing here:
- in the beginning, we use `testAccCheckNewUserGroup` that basically creates and user and adds it to a group, `group1`
- then we remove the user from the group using `testAccRemoveUserFromGroup`
- then we check the user resource -- more specific we check that the `group_ids` set contains exactly one element, but I am not sure about this -- shouldn't it be 0 since we previously removed the user from the group?
- same for the data source

In order to check the functionality for this fix, I suggest leaving the test without the resource and data source checks, something like:
```
	{
				Config:             testAccCheckNewUserGroup,
				ExpectNonEmptyPlan: true,
				Check: resource.ComposeTestCheckFunc(
				testAccRemoveUserFromGroup(constant.GroupResource+".group1", constant.UserResource+"."+constant.NewUserResource),
			},
```
The indentation and syntax may not be ok, but the idea is that we want to see that a diff is generated when running this step:
- we create the user and add it to a group using `testAccCheckNewUserGroup` configuration
- we then remove the user from the group using `testAccRemoveUserFromGroup`
- after each step, a `terraform plan` is applied, in this case, the `terraform plan` will generate a diff (in the configuration the user appears to belong to a group, but when checking the remote state, the user doesn't belong to a group), but this is expected and because of `ExpectNonEmptyPlan: true` the test will pass.

ionoscloud/resource_user.go
```diff
@@ -377,5 +377,29 @@ func setUserData(d *schema.ResourceData, user *ionoscloud.User) error {
 		}
 	}
 
-	return nil
+	return d.Set("group_ids", getUserGroups(user))
```
adeatcu-ionos, line 380: @digna-ionos let's keep the consistency and let's make the set inside of the function, the function should end with a `return nil` sentence. Is not a big deal, but we have this format for other resources so we should keep it the same way.

adeatcu-ionos (changes requested):

ionoscloud/resource_user_test.go
adeatcu-ionos: @digna-ionos forgot to mention, to make sure that this is ok, run the suggested configuration without `ExpectNonEmptyPlan: true`, the test should fail since `terraform plan` will suggest a diff. Then run it using `ExpectNonEmptyPlan: true` and the test should pass.

ionoscloud/resource_user.go
```diff
@@ -162,11 +162,11 @@ func resourceUserCreate(ctx context.Context, d *schema.ResourceData, meta interf
 func resourceUserRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(services.SdkBundle).CloudApiClient
 
-	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, d.Id()).Execute()
+	user, apiResponse, err := client.UserManagementApi.UmUsersFindById(ctx, d.Id()).Depth(1).Execute()
```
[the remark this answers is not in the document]
digna-ionos: correct

adeatcu-ionos (changes requested):

adeatcu-ionos (approved):

PR #594, opened 2024-07-02
feat: api gateway route resource

glimberea:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/data_source_apigateway_route.go
+++ b/ionoscloud/data_source_apigateway_route.go
@@ -111,7 +111,7 @@ func dataSourceAPIGatewayRouteRead(ctx context.Context, d *schema.ResourceData, 
 	idValue, idOk := d.GetOk("id")
 	nameValue, nameOk := d.GetOk("name")
 	partialMatch := d.Get("partial_match").(bool)
-	gatewayId := d.Get("gateway_id").(string)
+	gatewayID := d.Get("gateway_id").(string)
 
 	id := idValue.(string)
 	name := nameValue.(string)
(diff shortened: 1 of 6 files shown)


ionoscloud/data_source_apigateway_route.go
```diff
@@ -11,24 +11,26 @@ import (
...
 		Schema: map[string]*schema.Schema{
 			"id": {
 				Type:        schema.TypeString,
-				Description: "The ID (UUID) of the API Gateway route.",
+				Description: "The ID (UUID) of the API Gateway Route.",
 				Optional:    true,
+				Computed:    true,
 			},
-			"apigateway_id": {
+			"gateway_id": {
```
avirtopeanu-ionos, line 24: Just a note that I agree with this change and I'll do this for the Gateway resource too

services/apigateway/route.go
```diff
@@ -78,9 +138,64 @@ func (c *Client) SetApiGatewayRouteRead(d *schema.ResourceData, route apigateway
...
+
+func setRouteConfig(d *schema.ResourceData) *apigateway.Route {
+	routeName := d.Get("name").(string)
+	routeType := d.Get("type").(string)
+	pathsRaw := d.Get("paths").([]interface{})
+	methodsRaw := d.Get("methods").([]interface{})
+	websocket := d.Get("websocket").(bool)
+	upstreamsRaw := d.Get("upstreams").([]interface{})
+
+	paths := make([]string, 0)
```
avirtopeanu-ionos, line 167: just a nitpick about these `make([]string, 0)` declarations,

If you insist in calling a `make` beforehand we should probably do `paths := make([]string, len(pathsRaw))`, otherwise simply remove the `make` call since it's redundant with the `appends` in the for loop

avirtopeanu-ionos (approved):

services/apigateway/route.go
glimberea: Fair, I'll change it to `make([]string, len(pathsRaw))`. This was more trying to figure out if this was what the linter wanted :))))

PR #612, opened 2024-07-24
fix: make id for ionoscloud_mongo_template computed

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/data_source_dbaas_mongo_template.go
+++ b/ionoscloud/data_source_dbaas_mongo_template.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	mongo "github.com/ionos-cloud/sdk-go-dbaas-mongo"
+
 	dbaasService "github.com/ionos-cloud/terraform-provider-ionoscloud/v6/services/dbaas"
 )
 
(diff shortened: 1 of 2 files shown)


ionoscloud/data_source_dbaas_mongo_template.go
```diff
@@ -24,11 +25,13 @@ func dataSourceDbassMongoTemplate() *schema.Resource {
 				Description:      "The unique ID of the template.",
 				ValidateDiagFunc: validation.ToDiagFunc(validation.IsUUID),
 				Optional:         true,
+				Computed:         true,
```
adeatcu-ionos, line 28: @cristiGuranIonos for many data sources that have the feature to retrieve using both `ID` and the `name` we have `ID` and `name` marked as `optional` but not as `computed`. Do we need to make this change for all the data sources?

ionoscloud/data_source_dbaas_mongo_template.go
```diff
@@ -94,6 +97,10 @@ func dataSourceDbassMongoTemplateRead(ctx context.Context, d *schema.ResourceDat
 	} else if len(templates) > 1 {
 		return diag.FromErr(fmt.Errorf("more than one DBaaS Mongo Template found for the specified search criteria"))
 	}
+
+	if err := d.Set("id", *templates[0].Id); err != nil {
```
adeatcu-ionos, line 101: @cristiGuranIonos I see that in `SetMongoDBTemplateData()` call from below we are setting the ID, why are we adding this other set here?

cristiGuranIonos: This is not something that I could reproduce. So I'm covering both bases, making is computed and setting it explicitly

ionoscloud/data_source_dbaas_mongo_template.go
cristiGuranIonos: might be, but we can wait until someone raises an issue

ionoscloud/data_source_dbaas_mongo_template.go
adeatcu-ionos: @cristiGuranIonos I am not sure that this will solve the problem but we will see.

adeatcu-ionos (approved):

PR #630, opened 2024-08-05
fix: apigateway minor doc fixes

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/resources/apigateway.md
+++ b/docs/resources/apigateway.md
@@ -16,7 +16,6 @@ An API gateway consists of the generic rules and configurations.
 ```
 resource "ionoscloud_apigateway" "example" {
     name              = "example-gateway"
-    logs              = true
     metrics           = true
     
     custom_domains {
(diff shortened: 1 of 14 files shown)


docs/resources/apigateway.md
```diff
@@ -16,7 +16,6 @@ An API gateway consists of the generic rules and configurations.
 ```
 resource "ionoscloud_apigateway" "example" {
     name              = "example-gateway"
-    logs              = true
```
glimberea, line 19: maybe we should leave `logs` in the example, just have them set to false?

glimberea (approved):

docs/resources/apigateway.md
cristiGuranIonos: it's set to false by default anyway, so I don't think it would add anything to the example. I can add it if you think it helps.

glimberea: I don't have strong feelings about it, so let's just leave it as it is then

Issue #641, opened 2024-08-07
ionoscloud_backup_unit Fail to deploy


jserranomarin:
## Description
Deployment of ionoscloud_bakcup_unit fail to deploy through terraform
The backup unit is created in the panel but terraform stay on "still creating" until timeout
After the time out or cancel after the terraform apply already deploy it; if you  make a new deployment or update the backup unit is label as "Tainted" and is replaced with another timeout
Import of a backup unit from terraform when it is deploy manually in the panel dont work either because it is labeled as tainted forcing a new deployment and timeout again

## Environment
Arsys Datacenter Designer
DE/TXL

Terraform version: 
Terraform v1.8.4

Provider version: 

hasicorp/random v3.6.2
ionoscloud versión tested:
      #version = "= 6.4.13"
      #version = ">= 6.4.17"
      #version = ">= 6.4.19"
      version = ">= 6.5"

## Additional Notes
This deployment works before several months ago
Could be related with the acronis platform which I think is new; 
Could be related with the Arsys's Tenant of ionos cloud

adeatcu-ionos:
Hello! I will take a look.

PR #643, opened 2024-08-08
fix: Implement workaround for the BackupUnits API bug

adeatcu-ionos:

Fixes https://github.com/ionos-cloud/terraform-provider-ionoscloud/issues/641

There is a problem with the BackupUnits API:
- `GET` on all the BackupUnits works properly;
- `GET` on `BackupUnits/BackupUnitID` doesn't work properly (the state in the response is `BUSY` and some fields are not populated properly)

Because of that, the client method `BackupunitsFindById` doesn't work properly.

I will open a bug for the API team, but until then I implemented a workaround: I created a method that simulates the `BackupunitsFindById` client method. The difference is that this method fetches all the backup units and then returns the searched one using ID matching.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_backup_unit.go
+++ b/ionoscloud/resource_backup_unit.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"strings"
 	"time"
 
@@ -97,7 +98,7 @@ func resourceBackupUnitRead(ctx context.Context, d *schema.ResourceData, meta in
 
 	client := meta.(services.SdkBundle).CloudApiClient
 
-	backupUnit, apiResponse, err := client.BackupUnitsApi.BackupunitsFindById(ctx, d.Id()).Execute()
+	backupUnit, apiResponse, err := BackupUnitFindByID(ctx, d.Id(), client)
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
@@ -252,7 +253,7 @@ func resourceBackupUnitImport(ctx context.Context, d *schema.ResourceData, meta 
 
 	buId := d.Id()
 
-	backupUnit, apiResponse, err := client.BackupUnitsApi.BackupunitsFindById(ctx, d.Id()).Execute()
+	backupUnit, apiResponse, err := BackupUnitFindByID(ctx, d.Id(), client)
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
@@ -280,7 +281,7 @@ func resourceBackupUnitImport(ctx context.Context, d *schema.ResourceData, meta 
 }
 
 func backupUnitReady(client *ionoscloud.APIClient, d *schema.ResourceData, c context.Context) (bool, error) {
-	backupUnit, apiResponse, err := client.BackupUnitsApi.BackupunitsFindById(c, d.Id()).Execute()
+	backupUnit, apiResponse, err := BackupUnitFindByID(c, d.Id(), client)
 	logApiRequestTime(apiResponse)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cristiGuranIonos:
Changelog and link the issue with a  fixes #number

ionoscloud/resource_backup_unit.go
```diff
@@ -346,3 +347,33 @@ func setBackupUnitData(d *schema.ResourceData, backupUnit *ionoscloud.BackupUnit
...
+	for _, backupUnit := range *backupUnits.Items {
+		if backupUnit.Id == nil {
+			return backupUnit, apiResponse, fmt.Errorf("expected a backup unit with a valid ID but received 'nil' instead")
+		}
+		if *backupUnit.Id == backupUnitID {
+			return backupUnit, apiResponse, nil
+		}
+	}
+	// If the backup unit wasn't found, simulate a 404 error.
+	apiResponse.StatusCode = http.StatusNotFound
```
avirtopeanu-ionos, line 377: it's weird to mutate the API Response's status code. The rest of the API Response will not reflect a 404 error. For example the Status string, etc.

This object will now have fields populated like an OK response, except the StatusCode will be 404

ionoscloud/resource_backup_unit.go
```diff
@@ -346,3 +347,33 @@ func setBackupUnitData(d *schema.ResourceData, backupUnit *ionoscloud.BackupUnit
...
 	return nil
 }
+
+// BackupUnitFindByID it's a custom function that returns a backup unit using its ID. The difference
+// between this function and the client BackupUnitsApi.BackupunitsFindById is that this function
+// retrieves the whole list of backup units and then iterates over them, returning the backup unit
+// when the ID is matched. This function was implemented because the response values in
+// the "BackupUnitsApi.BackupunitsFindById" method are invalid, but the values in
+// "BackupUnitsApi.BackupunitsGet" method are valid. This is a workaround for the API problem with
+// invalid values in the response for BackupUnitsApi.BackupunitsFindById.
```
avirtopeanu-ionos, line 357: (nitpick) This is well detailed and explains all the context but honestly this could be simplified, like

```
BackupUnitFindByID simulates a FindByID function by filtering backup units from BackupunitsGet using the given ID.
This is done because of a temporary bug in the API with the regular FindByID function
```

avirtopeanu-ionos (approved):

ionoscloud/resource_backup_unit.go
avirtopeanu-ionos: (comments should explain WHY it exists, the code in the function itself is readable and people can tell that the whole list of backups is returned and iterated over etc)

ionoscloud/resource_backup_unit.go
adeatcu-ionos: @avirtopeanu-ionos `it's weird to mutate the API Response's status code.` I agree, I made this change since the status code it's the only thing we look at when deciding if the backup unit it's still there. I'm open to suggestions to avoid this weird mutation.

ionoscloud/resource_backup_unit.go
adeatcu-ionos: @avirtopeanu-ionos I agree, I'll make the change, your suggestion is clearer.

adeatcu-ionos: @avirtopeanu-ionos done in e4bb248a188c64025c74ce34b42a02d51ef8020f

ionoscloud/resource_backup_unit.go
cristiGuranIonos: we either mutate the apiresponse, or we check the error with error.is.

cristiGuranIonos: let's add a todo in the function comment that we need to replace the function once the api fixes the findByID.  This being a temporary fix, it doesn;t matter

adeatcu-ionos: @cristiGuranIonos done

cristiGuranIonos (approved):

rmocanu-ionos (approved):

PR #693, opened 2024-10-18
feat: Implement 'maintenance_window', 'tier' and multiple endpoints f…

adeatcu-ionos:
…or VPN resources

## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [ ] Jira task updated


ionoscloud/data_source_vpn_ipsec_gateway.go
```diff
@@ -79,6 +79,28 @@ func dataSourceVpnIPSecGateway() *schema.Resource {
 				Description: "The IKE version that is permitted for the VPN tunnels.",
 				Computed:    true,
 			},
+			"maintenance_window": {
```
digna-ionos, line 82: Is it possible to have multiple maintance_window objects?

adeatcu-ionos: @digna-ionos no, it is not possible. For resources we use `MaxItems: 1` to restrict the creation, but for data sources there is no need to define `MaxItems: 1` since in the read function only one `maintenance_window` will be read.

digna-ionos (dismissed):

services/vpn/client.go
```diff
@@ -20,14 +20,13 @@ type Client struct {
 
 var (
 	// AvailableLocations is a list of supported locations for VPN
-	AvailableLocations = []string{"de/fra", "de/txl"}
-	// todo for now we only support de/fra and de/txl.
-	// AvailableLocations = []string{"de/fra", "de/txl", "es/vit", "gb/lhr", "us/ewr", "us/las", "us/mci", "fr/par"}
+	AvailableLocations = []string{"de/fra", "de/txl", "es/vit", "gb/bhx", "gb/lhr", "us/ewr", "us/las", "us/mci", "fr/par"}
```
cristiGuranIonos, line 23: We should probably add gb/bhx to docs

docs/resources/vpn_ipsec_gateway.md
```diff
@@ -143,6 +148,10 @@ resource "ionoscloud_vpn_ipsec_gateway" "example" {
       Gateway. **Note**: this should be the subnet already assigned to the LAN
 * `version` - (Required)[string] The IKE version that is permitted for the VPN tunnels. Default: `IKEv2`. Possible
   values: `IKEv2`.
+* `maintenance_window` - (Optional)(Computed)[string] A weekly 4 hour-long window, during which maintenance might occur.
```
cristiGuranIonos, line 151: Not a string, a block or a list. We should fix that for mongo and psql too.

ionoscloud/resource_vpn_ipsec_gateway.go
```diff
@@ -50,8 +50,9 @@ func resourceVpnIPSecGateway() *schema.Resource {
 				Type:        schema.TypeList,
 				Description: "The network connection for your gateway. Note: all connections must belong to the same datacenter.",
 				MinItems:    1,
-				MaxItems:    10,
-				Required:    true,
+				// TODO -- Change this from 10 to 5 or leave this validation for the API
```
cristiGuranIonos, line 53: do we need to change to 5?

ionoscloud/resource_vpn_ipsec_gateway.go
```diff
@@ -87,6 +88,33 @@ func resourceVpnIPSecGateway() *schema.Resource {
...
+						},
+						"day_of_the_week": {
+							Type:        schema.TypeString,
+							Description: "The name of the week day",
+							Required:    true,
+						},
+					},
+				},
+			},
+			"tier": {
```
cristiGuranIonos, line 112: if it has a default, we can just use that instead of computed as it's clearer on plan what value it will get created with.  I'm guessing STANDARD.

docs/resources/vpn_ipsec_gateway.md
adeatcu-ionos: @cristiGuranIonos thanks, good catch! I'll modify this.

ionoscloud/resource_vpn_ipsec_gateway.go
adeatcu-ionos: @cristiGuranIonos the `maxItems` it's now 5 in the swagger. I added this `TODO` but I'm not sure about this change, I think it's better to simply remove this validation and let the API do all the validation.

cristiGuranIonos (dismissed):

PR #697, opened 2024-10-25
feature: add hostname to server and vcpu_server and servers

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/data-sources/server.md
+++ b/docs/data-sources/server.md
@@ -44,6 +44,7 @@ data "ionoscloud_server" "example" {
 The following attributes are returned by the datasource:
 
 * `template_uuid` - The UUID of the template for creating a CUBE server; the available templates for CUBE servers can be found on the templates resource
+* `hostname` - The hostname of the resource. Allowed characters are a-z, 0-9 and - (minus). Hostname should not start with minus and should not be longer than 63 characters.
 * `id` - The id of that resource
 * `name` - The name of that resource
 * `type` - Server usages: [ENTERPRISE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/virtual-servers) or [CUBE](https://docs.ionos.com/cloud/compute-engine/virtual-servers/cloud-cubes)
--- a/docs/resources/volume.md
+++ b/docs/resources/volume.md
@@ -93,7 +93,7 @@ resource "ionoscloud_volume" "example" {
   user_data               = "foo"
 }
 
-resource "ionoscloud_volume" "example" {
+resource "ionoscloud_volume" "example2" {
   datacenter_id           = ionoscloud_datacenter.example.id
   server_id               = ionoscloud_server.example.id
   name                    = "Another Volume Example"
--- a/ionoscloud/data_source_servers.go
+++ b/ionoscloud/data_source_servers.go
@@ -47,7 +47,10 @@ func dataSourceServers() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
-
+						"hostname": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 						"cores": {
 							Type:     schema.TypeInt,
 							Computed: true,
@@ -368,6 +371,7 @@ func SetServerProperties(server ionoscloud.Server) map[string]interface{} {
 	if server.Properties != nil {
 		utils.SetPropWithNilCheck(serverMap, "template_uuid", server.Properties.TemplateUuid)
 		utils.SetPropWithNilCheck(serverMap, "name", server.Properties.Name)
+		utils.SetPropWithNilCheck(serverMap, "hostname", server.Properties.Hostname)
 		utils.SetPropWithNilCheck(serverMap, "cores", server.Properties.Cores)
 		utils.SetPropWithNilCheck(serverMap, "ram", server.Properties.Ram)
 		utils.SetPropWithNilCheck(serverMap, "availability_zone", server.Properties.AvailabilityZone)
--- a/ionoscloud/data_source_servers.go
+++ b/ionoscloud/data_source_servers.go
@@ -49,7 +49,7 @@ func dataSourceServers() *schema.Resource {
 						},
 						"hostname": {
 							Type:     schema.TypeString,
-							Optional: true,
+							Computed: true,
 						},
 						"cores": {
 							Type:     schema.TypeInt,
--- a/ionoscloud/resource_server.go
+++ b/ionoscloud/resource_server.go
@@ -48,7 +48,7 @@ func resourceServer() *schema.Resource {
 			"hostname": {
 				Type:             schema.TypeString,
 				Optional:         true,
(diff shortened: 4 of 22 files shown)


docs/data-sources/server.md
```diff
@@ -44,6 +44,7 @@ data "ionoscloud_server" "example" {
 The following attributes are returned by the datasource:
 
 * `template_uuid` - The UUID of the template for creating a CUBE server; the available templates for CUBE servers can be found on the templates resource
+* `hostname` - The hostname of the resource. Allowed characters are a-z, 0-9 and - (minus). Hostname should not start with minus and should not be longer than 63 characters.
```
adeatcu-ionos, line 47:
```suggestion
* `hostname` - The hostname of the resource.
```
No need to specify the input format in the data source.

docs/resources/volume.md
```diff
@@ -93,7 +93,7 @@ resource "ionoscloud_volume" "example" {
   user_data               = "foo"
 }
 
-resource "ionoscloud_volume" "example" {
```
adeatcu-ionos, line 96: @cristiGuranIonos maybe we can add an example in the documentation for the hostname, but no problem if we leave it as it is, it is explained well enough in the docs.

ionoscloud/data_source_servers.go
```diff
@@ -47,7 +47,10 @@ func dataSourceServers() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
-
+						"hostname": {
+							Type:     schema.TypeString,
+							Optional: true,
```
adeatcu-ionos, line 52: @cristiGuranIonos shouldn't it be `Computed: true` here?

ionoscloud/resource_server.go
```diff
@@ -45,6 +45,12 @@ func resourceServer() *schema.Resource {
 				Required:         true,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 			},
+			"hostname": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Description:      "The hostname of the resource. Allowed characters are a-z, 0-9 and - (minus). Hostname should not start with minus and should not be longer than 63 characters.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.All(validation.StringIsNotWhiteSpace, validation.StringLenBetween(1, 63))),
```
adeatcu-ionos, line 52: @cristiGuranIonos same discussion about validation. Here we check the length of the string, we are not checking the content of the string, basically we do a partial check. My suggestion: remove this `ValidateDiagFunc` and let the API do all the validation.

ionoscloud/resource_server.go
```diff
@@ -1373,7 +1384,12 @@ func getServerData(d *schema.ResourceData) (*ionoscloud.Server, error) {
 			server.Properties.CpuFamily = &vStr
 		}
 	}
-
+	if v, ok := d.GetOk("hostname"); ok {
+		if v.(string) != "" {
```
adeatcu-ionos, line 1388: @cristiGuranIonos I see that we have empty string checks for some fields here.

With the current code (with validations) this will never be the case, the string will never be empty.

If you choose the remove the validation as I suggested in a comment above and also remove this empty string check, the request will be populated with an empty string `hostname` and the API request will fail with a validation error, there is nothing wrong with that.

With or without validation, I suggest removing this check, not sure how it helps.

ionoscloud/resource_server_test.go
```diff
@@ -68,6 +69,7 @@ func TestAccServerBasic(t *testing.T) {
 				Config: testAccCheckServerNoNicUpdate,
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(constant.ServerResource+"."+constant.ServerTestResource, "name", constant.ServerTestResource),
+					resource.TestCheckResourceAttr(constant.ServerResource+"."+constant.ServerTestResource, "hostname", constant.ServerTestHostname),
```
adeatcu-ionos, line 72: @cristiGuranIonos since this is the configuration for the update: `testAccCheckServerNoNicUpdate` I suggest using a new `hostname` value for this test, maybe something like `myhostnameupdated`.

ionoscloud/resource_vcpu_server.go
```diff
@@ -27,6 +27,12 @@ func resourceVCPUServer() *schema.Resource {
 				Required:         true,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 			},
+			"hostname": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Description:      "The hostname of the resource. Allowed characters are a-z, 0-9 and - (minus). Hostname should not start with minus and should not be longer than 63 characters.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.All(validation.StringIsNotWhiteSpace, validation.StringLenBetween(1, 63))),
```
adeatcu-ionos, line 34: @cristiGuranIonos same comment as the one from above related to validation.

ionoscloud/resource_vcpu_server_test.go
```diff
@@ -109,6 +110,7 @@ func TestAccServerVCPUBasic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckServerVCPUExists(constant.ServerVCPUResource+"."+constant.ServerTestResource, &server),
 					resource.TestCheckResourceAttr(constant.ServerVCPUResource+"."+constant.ServerTestResource, "name", constant.ServerTestResource),
+					resource.TestCheckResourceAttr(constant.ServerVCPUResource+"."+constant.ServerTestResource, "hostname", constant.ServerTestHostname),
```
adeatcu-ionos, line 113: @cristiGuranIonos same, maybe add an update test for the `hostname` value.

adeatcu-ionos (changes requested):

ionoscloud/resource_vcpu_server.go
cristiGuranIonos: will keep it, it doesn't hurt for this field as it's optional

docs/resources/volume.md
cristiGuranIonos: we'll leave it as it is for now. should be clear enough how it's used.

ionoscloud/resource_server.go
cristiGuranIonos: will leave the partial check, doesn't hurt anything  even if the validation changes for this.

ionoscloud/resource_server_test.go
cristiGuranIonos: added, value only "updated"

ionoscloud/resource_vcpu_server_test.go
cristiGuranIonos: added

adeatcu-ionos (approved):

PR #729, opened 2024-12-18
feat: Implement Monitoring Service

adeatcu-ionos:
## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [ ] Jira task updated


adeatcu-ionos:
**Some notes about this PR:**
- The structure may not be the best, in terms of where to store the service, the resources/data-sources, etc but I followed the bucket implementation because I just wanted to integrate the service and make it work. I tried to change the structure but I realized that it's a lot of work and it was out of scope for this PR. I will leave this as it is and we will get back to it later.
- The sentence above also applies for the tests. Ideally, for tests, we would need to make a provider instance so we can reference all the clients that we usually instantiate inside a provider.
- The tests are not done yet, I still need to add some more
- The SDK is generated locally, I will make a bundle release soon and then I will modify the PR accordingly.

rmocanu-ionos (approved):

PR #749, opened 2025-01-27
feat: move postgresql to sdk-go-bundle

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


services/dbaas/database.go
```diff
@@ -36,17 +34,15 @@ func (c *PsqlClient) GetDatabases(ctx context.Context, clusterId string) (pgsql.
...
 	resourceName := "PgSQL database"
-	d.SetId(*database.Id)
-	if database.Properties == nil {
-		return fmt.Errorf("expected properties in the response for the PgSql database with ID: %s, but received 'nil' instead", *database.Id)
-	}
-	if database.Properties.Name != nil {
-		if err := d.Set("name", *database.Properties.Name); err != nil {
+	d.SetId(database.Id)
+
+	if database.Properties.Name != "" {
```
adeatcu-ionos, line 39: @cristiGuranIonos do we need this check here? I don't think so.

services/dbaas/users.go
```diff
@@ -182,12 +182,10 @@ func SetUserMongoData(d *schema.ResourceData, user *mongo.User) error {
...
 	resourceName := "PgSQL user"
-	d.SetId(*user.Id)
-	if user.Properties == nil {
-		return fmt.Errorf("expected properties in the response for the PgSql user with ID: %s, but received 'nil' instead", *user.Id)
-	}
-	if user.Properties.Username != nil {
-		if err := d.Set("username", *user.Properties.Username); err != nil {
+	d.SetId(user.Id)
+
+	if user.Properties.Username != "" {
```
adeatcu-ionos, line 187: @cristiGuranIonos same here, I don't think we need this check.

adeatcu-ionos (changes requested):

ionoscloud/data_source_dbaas_pgsql_cluster.go
```diff
@@ -200,8 +200,8 @@ func dataSourceDbaasPgSqlReadCluster(ctx context.Context, d *schema.ResourceData
 
 		var results []dbaas.ClusterResponse
 
-		if clusters.Items != nil && len(*clusters.Items) > 0 {
-			for _, clusterItem := range *clusters.Items {
+		if len(clusters.Items) > 0 {
```
glimberea, line 203: why not just remove this check? it's a little redundant since the for loop will be skipped either way if `len(items)` is 0

services/dbaas/database.go
adeatcu-ionos: @adeatcu-ionos not sure if this was the case when I left the comment and I didn't check properly, but we need this check here because `Name` is a pointer.

services/dbaas/users.go
adeatcu-ionos: @adeatcu-ionos not sure if this was the case when I left the comment and I didn't check properly, but we need this check here because `Name` is a pointer.

services/dbaas/database.go
adeatcu-ionos: @adeatcu-ionos I still don't know why we have these checks here but I will leave them as they are because maybe I am missing some strange edge case.

glimberea (approved):

glimberea (approved):

adeatcu-ionos (approved):

adeatcu-ionos (approved):

PR #801, opened 2025-03-27
fix: save password for ionoscloud_container_registry_token in passwor…

### LINKED ISSUE #799: Attribute access to actual token of ionoslcoud_container_registry_token resource
### Current Provider Version

```
6.7.3
```

### Use-cases

Create individual imagePullSecrets per Kubernetes namespace with individual access scopes to the IONOS registry.

### Attempted Solutions

Creating tokens via the `ionoscloud_container_registry_token` resource and writing them to the cluster via the `kubernetes_secret` resource. However, the token can only be written to a file, and it's unclear how to use that in other places or use with remote state storage.

### Proposal

Make the actual token a computed attribute of the newly created `ionoscloud_container_registry_token` resource.

On import, this attribute would have to be empty and could only be filled using a re-create (e.g. by tainting the resource).

cristiGuranIonos:
…d field of resource
Fixes #799
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_container_registry_token_test.go
+++ b/ionoscloud/resource_container_registry_token_test.go
@@ -45,6 +45,8 @@ func TestAccContainerRegistryTokenBasic(t *testing.T) {
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "scopes.0.type", "repository"),
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "status", "enabled"),
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "name", constant.ContainerRegistryTokenTestResource),
+					resource.TestCheckResourceAttrSet(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "credentials.0.password"),
+					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "credentials.0.username", constant.ContainerRegistryTokenTestResource),
 				),
 			},
 			{
@@ -104,7 +106,7 @@ func TestAccContainerRegistryTokenBasic(t *testing.T) {
 			},
 			{
 				Config:      getConfigurationFromTemplate(testAccDataSourceContainerRegistryTokenMultipleTokensFound, templateData),
-				ExpectError: regexp.MustCompile("more than one token found with the specified criteria: name = test-container-registry-token"),
+				ExpectError: regexp.MustCompile("more than one token found with the specified criteria: name = " + constant.ContainerRegistryTokenTestResource),
 			},
 		},
 	})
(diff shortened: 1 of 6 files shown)


rmocanu-ionos (approved):

services/containerregistry/registries.go
```diff
@@ -439,22 +429,24 @@ func SetTokenData(d *schema.ResourceData, tokenProps cr.TokenProperties) error {
...
-func SetCredentials(credentials cr.Credentials) map[string]interface{} {
+// SetCredentials sets the credentials for the registry token. username can be set on read, password will be set only on create
+func SetCredentials(credentials cr.Credentials) map[string]any {
 
-	credentialsEntry := map[string]interface{}{}
-
-	utils.SetPropWithNilCheck(credentialsEntry, "username", credentials.Username)
+	credentialsEntry := map[string]any{}
+	credentialsEntry["username"] = credentials.Username
+	if credentials.Password != "" {
```
adeatcu-ionos, line 437: @cristiGuranIonos this check is necessary because the `password` is returned only on `POST` requests, I assume that for `GET` requests, the password will be `""`, right?

services/containerregistry/registries.go
```diff
@@ -439,22 +429,24 @@ func SetTokenData(d *schema.ResourceData, tokenProps cr.TokenProperties) error {
...
+		credentialsEntry["password"] = credentials.Password
```
adeatcu-ionos, line 438: @cristiGuranIonos why we have both

```
credentialsEntry["password"] = credentials.Password
```

and 

```
utils.SetPropWithNilCheck(credentialsEntry, "password", credentials.Password)
```

I think the first one should be enough since the `password` will never be a pointer.

services/containerregistry/registries.go
cristiGuranIonos: yes, but the function is used in the data source which does not have the `password` field and would crash on write.

ionoscloud/resource_container_registry_token_test.go
```diff
@@ -45,6 +45,8 @@ func TestAccContainerRegistryTokenBasic(t *testing.T) {
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "scopes.0.type", "repository"),
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "status", "enabled"),
 					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "name", constant.ContainerRegistryTokenTestResource),
+					resource.TestCheckResourceAttrSet(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "credentials.0.password"),
+					resource.TestCheckResourceAttr(constant.ContainerRegistryTokenResource+"."+constant.ContainerRegistryTokenTestResource, "credentials.0.username", constant.ContainerRegistryTokenTestResource),
```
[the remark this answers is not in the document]
cristiGuranIonos: yes, it should be the same

adeatcu-ionos:
@cristiGuranIonos also, we talked about the tests because for the CR tokens the 8/8 step was failing, did you manage to fix that one? Asking because if not, we need to create an issue to solve it.

adeatcu-ionos (approved):

cristiGuranIonos:
> @cristiGuranIonos also, we talked about the tests because for the CR tokens the 8/8 step was failing, did you manage to fix that one? Asking because if not, we need to create an issue to solve it.

Yes, fixed by adding the correct name in the err check

PR #802, opened 2025-03-27
Feat: add server type to k8s nodepool and set cpu_family to optional

glimberea:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


ionoscloud/resource_k8s_node_pool.go
```diff
@@ -144,10 +144,17 @@ func resourceK8sNodePool() *schema.Resource {
...
-				Required:         true,
+				Optional:         true,
 				ForceNew:         true,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 			},
+			"server_type": {
+				Type:             schema.TypeString,
+				Description:      "The server type for the compute engine",
+				Optional:         true,
+				Default:          "DedicatedCore",
```
adeatcu-ionos, line 155: @glimberea we decided in the past that we should not have default values for the attributes anymore. The attributes are easier to maintain this way, if the default value changes in the API, we also need to change the value in Terraform. Instead, I think you should mark this as `Computed: true` since I assume that if `server_type` is missing, the API will set a value for it.

cristiGuranIonos:
changelog

ionoscloud/resource_k8s_node_pool.go
```diff
@@ -144,10 +144,17 @@ func resourceK8sNodePool() *schema.Resource {
...
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"DedicatedCore", "VCPU"}, true)),
```
adeatcu-ionos, line 156: @glimberea not sure if we need to ignore case here. I see the enum in the swagger and I suppose that if we send `dedicatedCORE` it should fail, I'm not sure if an exact match is done in the API, so please double check this.

ionoscloud/resource_k8s_node_pool.go
```diff
@@ -144,10 +144,17 @@ func resourceK8sNodePool() *schema.Resource {
 			"cpu_family": {
 				Type:             schema.TypeString,
 				Description:      "CPU Family",
-				Required:         true,
+				Optional:         true,
```
adeatcu-ionos, line 147: @glimberea what happens if `cpu_family` is absent from the configuration? A default value will be set? If so, we should mark this as `Computed: true`.

ionoscloud/resource_k8s_node_pool.go
```diff
@@ -144,10 +144,17 @@ func resourceK8sNodePool() *schema.Resource {
...
 				ForceNew:         true,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 			},
+			"server_type": {
+				Type:             schema.TypeString,
+				Description:      "The server type for the compute engine",
+				Optional:         true,
```
adeatcu-ionos, line 154: @glimberea please double check if `ForceNew: true` is required here.

ionoscloud/resource_k8s_node_pool_test.go
```diff
@@ -38,7 +38,7 @@ func TestAccK8sNodePoolBasic(t *testing.T) {
 					resource.TestCheckResourceAttr(constant.ResourceNameK8sNodePool, "maintenance_window.0.day_of_the_week", "Monday"),
 					resource.TestCheckResourceAttr(constant.ResourceNameK8sNodePool, "maintenance_window.0.time", "09:00:00Z"),
 					resource.TestCheckResourceAttr(constant.ResourceNameK8sNodePool, "auto_scaling.#", "0"),
-					resource.TestCheckResourceAttr(constant.ResourceNameK8sNodePool, "cpu_family", "INTEL_XEON"),
+					resource.TestCheckResourceAttr(constant.ResourceNameK8sNodePool, "server_type", "DedicatedCore"),
```
adeatcu-ionos, line 41: @glimberea I see that we completely replaced `cpu_family` with `server_type`. This means that we are not testing the creation with `cpu_family` in the configuration anymore. Instead, since `cpu_family` and `server_type` are both optional, I suggest the following test cases:
1. Create node pool when both attributes are missing from the configuration - in this case, for the `computed` attributes we need to check if the attributes are properly set in state
2. Create node pool when both attributes are in the configuration - in this case we need to check that the values set in state are the ones from the configuration
3. Any update/removal for these optional fields - we need to test if we can successfully remove/change optional fields, but I think this applies only for `server_type` since `cpu_family` is marked with `ForceNew: true`.

adeatcu-ionos (changes requested):

ionoscloud/resource_k8s_node_pool.go
glimberea: I had a discussion about this with @cristiGuranIonos and we thought it would be better to have the default value, because we want to catch changes made outside of the terraform configuration (e.g. through DCD), not automatically adopt them if the attribute is not explicitly set.

ionoscloud/resource_k8s_node_pool.go
glimberea: it does not, it leaves it empty

ionoscloud/resource_k8s_node_pool_test.go
adeatcu-ionos: @glimberea 

> since server_type has a default value and can be considered always set. 

I disagree, it's important to have a test configuration in which `server_type` is missing. We want to test the correct behavior so we want to test that the default value is successfully set when `server_type` is missing from the configuration.

adeatcu-ionos (approved):

PR #833, opened 2025-06-11
fix: ionoscloud_lan should retry if lan is delete-protected

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/ionoscloud/resource_lan.go
+++ b/ionoscloud/resource_lan.go
@@ -257,7 +257,7 @@ func resourceLanDelete(ctx context.Context, d *schema.ResourceData, meta interfa
 	err := retry.RetryContext(ctx, d.Timeout(schema.TimeoutDelete), func() *retry.RetryError {
 		apiResponse, err := client.LANsApi.DatacentersLansDelete(ctx, dcId, d.Id()).Execute()
 		if err != nil {
-			if apiResponse.Response != nil && apiResponse.StatusCode == 403 && strings.Contains(err.Error(), "is delete-protected by") {
+			if apiResponse != nil && apiResponse.Response != nil && apiResponse.StatusCode == 403 && strings.Contains(err.Error(), "is delete-protected by") {
 				log.Printf("[INFO] LAN %s is delete-protected, keep trying", d.Id())
 				return retry.RetryableError(fmt.Errorf("lan %s is delete-protected, keep trying %w", d.Id(), err))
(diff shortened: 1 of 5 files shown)


ionoscloud/resource_lan.go
```diff
@@ -252,13 +254,24 @@ func resourceLanDelete(ctx context.Context, d *schema.ResourceData, meta interfa
...
+				log.Printf("[INFO] LAN %s is delete-protected, keep trying", d.Id())
+				return retry.RetryableError(fmt.Errorf("lan %s is delete-protected, keep trying %w", d.Id(), err))
+			}
+			if httpNotFound(apiResponse) {
+				log.Printf("[INFO] LAN already deleted %s", d.Id())
+				return nil
+			}
+			return retry.NonRetryableError(fmt.Errorf("an error occurred while deleting lan dcId %s ID %s %w", dcId, d.Id(), err))
+		}
+		log.Printf("[DEBUG] resource with id %s still has not been deleted", d.Id())
```
adeatcu-ionos, line 270: @cristiGuranIonos I think that if we want to log a debug message here, it should be something like:
```suggestion
		log.Printf("[DEBUG] deletion started for LAN with ID: %v", d.Id())
```

ionoscloud/resource_lan.go
```diff
@@ -252,13 +254,24 @@ func resourceLanDelete(ctx context.Context, d *schema.ResourceData, meta interfa
 		return diag.FromErr(err)
 	}
 
-	apiResponse, err := client.LANsApi.DatacentersLansDelete(ctx, dcId, d.Id()).Execute()
-	logApiRequestTime(apiResponse)
-
+	err := retry.RetryContext(ctx, d.Timeout(schema.TimeoutDelete), func() *retry.RetryError {
```
adeatcu-ionos, line 257: @cristiGuranIonos can we simulate a test for this situation or is it hard to reproduce?

adeatcu-ionos (changes requested):

ionoscloud/resource_lan.go
cristiGuranIonos: the tests that will pass for inmemorydb and nfs will mean this works. Probably only inmemorydb as nfs has another issue on delete also.

PR #870, opened 2025-09-19
feat: Add support for new S3 locations

adeatcu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/services/objectstorage/client.go
+++ b/services/objectstorage/client.go
@@ -64,27 +64,14 @@ func NewClient(clientOptions clientoptions.TerraformClientOptions, config *filec
 		}
 	}
 
+	if clientOptions.StorageOptions.Region == "" {
+		clientOptions.StorageOptions.Region = "eu-central-3"
+	}
+
+	// TODO -- replace with WithObjectStorage from sdk-go-bundle
 	cfg := shared.NewConfigurationFromOptions(clientOptions.ClientOptions)
 	signer := awsv4.NewSigner(credentials.NewStaticCredentials(clientOptions.StorageOptions.AccessKey, clientOptions.StorageOptions.SecretKey, ""))
-	cfg.MiddlewareWithError = func(r *http.Request) error {
-		var reader io.ReadSeeker
-		if r.Body != nil {
-			bodyBytes, err := io.ReadAll(r.Body)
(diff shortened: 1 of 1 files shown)


services/objectstorage/client.go
```diff
@@ -82,8 +93,57 @@ func NewClient(clientOptions clientoptions.TerraformClientOptions, config *filec
...
+)
+
+func (c *Client) ChangeConfigURL(bucketRegion string) error {
+	var region string
+
+	if c.globalRegion != "" && c.globalRegion != bucketRegion {
+		return fmt.Errorf("global region: %v is different from the bucket region: %v, if you want to use the bucket region, please unset the global region", c.globalRegion, bucketRegion)
+	}
+	region = bucketRegion
+	if region != "" && !slices.Contains(ValidS3Locations, region) {
```
cristiGuranIonos, line 122: will this be a problem if someone declares in the config file a custom location and sets a custom url for it?

services/objectstorage/client.go
```diff
@@ -82,8 +93,57 @@ func NewClient(clientOptions clientoptions.TerraformClientOptions, config *filec
...
+		return fmt.Errorf("invalid region: %v, available regions are: %s", region, strings.Join(ValidS3Locations, ","))
+	}
+
+	cfg := c.client.GetConfig()
+	cfg.Servers = shared.ServerConfigurations{
+		{
+			URL: regionToURL[region],
+		},
+	}
+	cfg.MiddlewareWithError = func(r *http.Request) error {
```
cristiGuranIonos, line 132: if this is the same in client init, it can probably be extracted to a separate function?

services/objectstorage/client.go
adeatcu-ionos: @cristiGuranIonos done in 81f2e753c7853ff8ddfc6fa7f7f7b44b734f6df1

services/objectstorage/client.go
adeatcu-ionos: @cristiGuranIonos the functions slightly differ and they reference variables that are visible in the same scope in which the function is defined and I also can't change the function signature.

Issue #891, opened 2025-11-04
Implement data_source_kafka_users


mjuergens-cg:
### Current Provider Version

```
170e5ee
```

### Use-cases

We would like to fully automate the process of setting up Kafka and configuring access of our applications to it. Thereby we need a way to gather the user configuration (keys / certificates) automatically.

### Attempted Solutions

Checked the API endpoint: https://api.ionos.com/docs/kafka/v1/#tag/Users/operation/clustersUsersAccessGet - this is supplying what we need.

### Proposal

Implement data_source_kafka_users in Terraform provider.

PR #919, opened 2026-01-19
feat: Add Kafka Users data sources

adeatcu-ionos:

Implement Kafka Users data sources & ephemeral resources: https://github.com/ionos-cloud/terraform-provider-ionoscloud/issues/891.

This implements the following entities:
- Kafka Users data source
- Kafka User Access Credentials data source
- Kafka User Access Credentials ephemeral resource - for cases in which the customer doesn't want to store sensitive data into the state.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


internal/framework/utils/validators/string_uuid.go
```diff
@@ -0,0 +1,35 @@
+package validators
```
cristiGuranIonos: I don't think we need the utils folder. We can have validators

PR #923, opened 2026-02-11
feat: add regional endpoints cloud

glimberea:
### Issue
 
Currently, Cloud API has only one global endpoint and resource localization is handled at the API level. Add support for location-based endpoints for Cloud API. Some resources are regional, while others are truly global. The new implementation should be backwards compatible.

### Initial idea

The original implementation used a common `CloudApiClient` for all operations with compute resources. Instead of using a common client, the first attempt was to use a common client `Configuration`, which will then be deep-copied and used to instantiate a new client for every operation. Deep-copying proved to add too much overhead to the client instantiation, so this approach was abandoned.

### Final approach

For all clients in the `SdkBundle` struct, `FileConfig` and `ClientOptions` are used as the base for client configuration. Instead of passing them as parameters to independent constructor functions for dedicated API `Client` structs, store them in the `SdkBundle` struct and add a client constructor method (`SdkBundle.NewCloudAPIClient`) to this struct. Each client created by this method will have its own config.

*TODO:* use new release of `sdk-go-bundle/shared` instead of commit

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/ionos-cloud/sdk-go-bundle/products/objectstoragemanagement/v2 v2.0.0
 	github.com/ionos-cloud/sdk-go-bundle/products/vmautoscaling/v2 v2.0.1
 	github.com/ionos-cloud/sdk-go-bundle/products/vpn/v2 v2.0.2
-	github.com/ionos-cloud/sdk-go-bundle/shared v0.1.7-0.20260209105343-d884c00c537c
+	github.com/ionos-cloud/sdk-go-bundle/shared v0.1.7
 	github.com/ionos-cloud/sdk-go/v6 v6.3.6
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/stretchr/testify v1.10.0
--- a/go.sum
+++ b/go.sum
@@ -148,8 +148,8 @@ github.com/ionos-cloud/sdk-go-bundle/products/vmautoscaling/v2 v2.0.1 h1:FOwVFj5
 github.com/ionos-cloud/sdk-go-bundle/products/vmautoscaling/v2 v2.0.1/go.mod h1:InNuTTzgLZker3i9JpyTho9mHr5DDIFnEPwykbSIfyw=
 github.com/ionos-cloud/sdk-go-bundle/products/vpn/v2 v2.0.2 h1:qR/ul88v/3ZF+ZLFFv6LE9TrjZS1qM4R10UlXCaQzcs=
 github.com/ionos-cloud/sdk-go-bundle/products/vpn/v2 v2.0.2/go.mod h1:V6WHbWsQDlZsLtHqgsXO81Z9eJBsczE3Q6VY8J+rutc=
-github.com/ionos-cloud/sdk-go-bundle/shared v0.1.7-0.20260209105343-d884c00c537c h1:3+qu3NTa/uK7VWhSAgHx3vUkxSWH16pgRUjFI2u/c0A=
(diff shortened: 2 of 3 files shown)


ionoscloud/utils.go
```diff
@@ -110,3 +110,38 @@ func httpNotFound(resp *ionoscloud.APIResponse) bool {
...
+	}
+}
+
+// validateImportIdentifiers checks that all resource IDs within the import identifier are non-empty, returning an error
+// if any of them is invalid. The importID parameter is used as context in the error message.
+func validateImportIdentifiers(importID string, resourceIDs []string) error {
+	for _, id := range resourceIDs {
+		if id == "" {
+			// TODO: is this error message good enough?
+			return fmt.Errorf("invalid identifier: resource IDs within the import identifier must all be non-empty, got %q", importID)
```
rmocanu-ionos: maybe something like this would be clearer?
return fmt.Errorf("invalid import identifier format: %q is missing required IDs (ensure all parts are provided and not empty)", importID)

ionoscloud/data_source_lan.go
```diff
@@ -109,6 +111,10 @@ func dataSourceLanRead(ctx context.Context, d *schema.ResourceData, meta interfa
 	var err error
 	var apiResponse *ionoscloud.APIResponse
 
+	location := d.Get("location").(string)
```
adeatcu-ionos: @glimberea please double check what happens if `location` is omitted from the config, it should be an empty string here, but let's double check, please.

ionoscloud/utils.go
```diff
@@ -110,3 +110,39 @@ func httpNotFound(resp *ionoscloud.APIResponse) bool {
...
+
+	return before, strings.Split(after, del)
+}
+
+// validateImportIDParts checks that all resource IDs within the import identifier are non-empty, returning an error
+// if any of them is invalid. The importID parameter is used as context in the error message.
+func validateImportIDParts(importID string, parts []string) error {
+	for _, id := range parts {
+		if id == "" {
+			return fmt.Errorf("invalid identifier: all parts of an import ID must be provided and non-empty, got %q", importID)
```
adeatcu-ionos:
```suggestion
			return fmt.Errorf("invalid identifier: all parts of an import ID must be provided and non-empty, got: %q", importID)
```

services/cloudapi/client.go
```diff
@@ -2,21 +2,119 @@ package cloudapi
...
-	"github.com/ionos-cloud/terraform-provider-ionoscloud/v6/utils/loadedconfig"
 )
 
+type Config struct {
+	fileConfig   *fileconfiguration.FileConfig
+	clientConfig *ionoscloud.Configuration
+}
+
+// NewConfig returns a new Config struct based on the provided client options and configuration file.
+func NewConfig(clientOptions clientoptions.TerraformClientOptions, fileConfig *fileconfiguration.FileConfig) *Config {
```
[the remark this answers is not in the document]
adeatcu-ionos: @glimberea please leave this comment as a general comment to this PR, otherwise it will be lost among the rest of the review comments. Please leave a comment in which you specify that we initially thought about copying the configuration / overwriting the configuration and some words on the new approach.

ionoscloud/data_source_snapshot.go
```diff
@@ -105,13 +105,16 @@ func dataSourceSnapshot() *schema.Resource {
...
-
 	id, idOk := d.GetOk("id")
 	name, nameOk := d.GetOk("name")
 	location, locationOk := d.GetOk("location")
 	size, sizeOk := d.GetOk("size")
 
+	client, err := meta.(bundleclient.SdkBundle).NewCloudAPIClient(location.(string))
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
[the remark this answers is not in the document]
glimberea: `d.Get` is a wrapper over `d.GetOk`, which ignores the existence check. It always returns the empty value for a parameter which is defined in the schema.

adeatcu-ionos (dismissed):

PR #979, opened 2026-05-06
Feat: add user bucket resource and data source

cristiGuranIonos:

User object bucket resource and data source.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


adeatcu-ionos:
@cristiGuranIonos please remove:

 ```User-owned buckets are not supported```
 
 From  `docs/resources/s3_bucket.md` and any other docs.

docs/resources/user_object_storage_bucket.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `name` - (Required)[string] The bucket name. Must be between 3 and 63 characters.
+- `region` - (Optional)[string] The region where the bucket is created. Defaults to `de` (Frankfurt). Available regions:
+  - `de` — Frankfurt, Germany (`https://s3.eu-central-1.ionoscloud.com`)
+  - `eu-central-2` — Berlin, Germany (`https://s3.eu-central-2.ionoscloud.com`)
+  - `eu-south-2` — Logroño, Spain (`https://s3.eu-south-2.ionoscloud.com`)
+- `object_lock_enabled` - (Optional)[bool] Whether Object Lock is enabled for the bucket. Defaults to `false`. **Cannot be changed after creation** — changing this value forces a new resource.
+- `force_destroy` - (Optional)[bool] Defaults to `false`. When set to `true`, all objects in the bucket are deleted before the bucket itself is destroyed, allowing Terraform to remove a non-empty bucket. **Use with caution** — this irreversibly deletes all bucket contents.
+- `timeouts` - (Optional) Timeouts for this resource.
+  - `create` - (Optional)[string] Time to wait for the bucket to be created. Default is `10m`.
```
adeatcu-ionos, line 50: `nitpick`: we can also add information here about the `read` and `update` timeouts, we are also supporting those.

internal/framework/provider/provider.go
```diff
@@ -16,12 +16,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+
```
adeatcu-ionos, line 18: is this newline intended here? we usually have imports in 3 sections but here we have 4 sections.

internal/framework/services/userobjectstorage/data_source_bucket.go
```diff
@@ -0,0 +1,98 @@
...
+// Configure wires the provider client.
+func (d *bucketDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
+	}
+	clientBundle, ok := req.ProviderData.(*bundleclient.SdkBundle)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Data Source Configure Type",
+			fmt.Sprintf("Expected *services.SdkBundle, got: %T. Please report this issue to the provider developers.", req.ProviderData),
```
adeatcu-ionos, line 74: `nitpick:`
```suggestion
			fmt.Sprintf("Expected *bundleclient.SdkBundle, got: %T. Please report this issue to the provider developers.", req.ProviderData),
```

internal/framework/services/userobjectstorage/data_source_bucket.go
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+
+	var data bucketDataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.Region.IsNull() || data.Region.IsUnknown() || data.Region.ValueString() == "" {
```
adeatcu-ionos:
```suggestion
	if data.Region.ValueString() == "" {
```

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+// NewBucketResource creates a new resource for the user object storage bucket.
+func NewBucketResource() resource.Resource {
+	return &bucketResource{}
+}
+
+type bucketResource struct {
+	client *userobjectstorage.Client
+}
+
+type bucketResourceModel struct {
```
adeatcu-ionos, line 41: `nitpick:` please sort these in alphabetical order.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+
+// Metadata sets the resource type name.
+func (r *bucketResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_user_object_storage_bucket"
+}
+
+// Schema returns the schema for the resource.
+func (r *bucketResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
```
adeatcu-ionos, line 59: `nitpick:` please order these attributes in alphabetical order.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+
+	createTimeout, diags := data.Timeouts.Create(ctx, utils.DefaultTimeout)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, createTimeout)
+	defer cancel()
+
+	if err := r.client.CreateBucket(ctx, data.Name, data.Region, data.ObjectLockEnabled, createTimeout); err != nil {
```
adeatcu-ionos:
```suggestion
	if err := r.client.CreateBucket(ctx, data, createTimeout); err != nil {
```

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+
+	var data bucketResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Region and object_lock_enabled cannot be read back from the API. Default them when
+	// missing, as happens after a plain `terraform import` by name only.
+	if data.Region.IsNull() || data.Region.IsUnknown() || data.Region.ValueString() == "" {
```
adeatcu-ionos:
```suggestion
	if data.Region.ValueString() == "" {
```

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+		return
+	}
+
+	var data bucketResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Region and object_lock_enabled cannot be read back from the API. Default them when
```
adeatcu-ionos: Why? Can't we make some additional requests in order to retrieve information about these?

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+	// missing, as happens after a plain `terraform import` by name only.
+	if data.Region.IsNull() || data.Region.IsUnknown() || data.Region.ValueString() == "" {
+		data.Region = types.StringValue(userobjectstorage.DefaultRegion)
+	}
+	if data.ObjectLockEnabled.IsNull() || data.ObjectLockEnabled.IsUnknown() {
```
adeatcu-ionos: `data.ObjectLockEnabled.ValueBool()` can be used here, it also handles the cases in which the `Bool` is `null` or `unknown`, case in which returns `false`.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	var state bucketResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	plan.ID = state.ID
```
adeatcu-ionos: `nitpick:` this is useless here, the `ID` will be the same.

The only case in which the `ID` will not be the same is when `name` is changed but if `name` is changed it won't reach this code anyways because of the `RequireReplace` set for the `name`.

With that being said, the function can be simplified to:
```
var plan bucketResourceModel
resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
if resp.Diagnostics.HasError() {
    return
}
resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+
+	deleteTimeout, diags := data.Timeouts.Delete(ctx, utils.DefaultTimeout)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, deleteTimeout)
+	defer cancel()
+
+	if err := r.client.DeleteBucket(ctx, data.Name, data.ForceDestroy, data.Region, deleteTimeout); err != nil {
```
adeatcu-ionos:
```suggestion
	if err := r.client.DeleteBucket(ctx, data, deleteTimeout); err != nil {
```

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+	// Region and object_lock_enabled cannot be read back from the API. Default them when
+	// missing, as happens after a plain `terraform import` by name only.
+	if data.Region.IsNull() || data.Region.IsUnknown() || data.Region.ValueString() == "" {
+		data.Region = types.StringValue(userobjectstorage.DefaultRegion)
+	}
+	if data.ObjectLockEnabled.IsNull() || data.ObjectLockEnabled.IsUnknown() {
+		data.ObjectLockEnabled = types.BoolValue(false)
+	}
+
+	found, err := r.client.GetBucket(ctx, data.Name, data.Region)
```
adeatcu-ionos: We are defining the `read` timeout for the schema but we are not using it in this function. If we decide not to use the `read` timeout, let's remove it from the schema.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// Update handles in-place changes (only force_destroy can change without replacement).
+func (r *bucketResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var plan bucketResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	var state bucketResourceModel
```
adeatcu-ionos: Since it doesn't make sense to use the `update` timeout, I suggest removing it from the schema.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,257 @@
...
+		return
+	}
+	if len(parts) == 2 {
+		region = parts[0]
+		name = parts[1]
+	} else {
+		name = parts[0]
+	}
+	if name == "" {
+		resp.Diagnostics.AddError("invalid bucket name", fmt.Sprintf("bucket name must not be empty. Got: %q", req.ID))
```
adeatcu-ionos: `nitpick:` since the name is empty I don't think it makes sense to print the value.

internal/framework/services/userobjectstorage/resource_bucket_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func testAccCheckUserBucketDestroy(s *terraform.State) error {
+	client := acctest.NewTestBundleClientFromEnv().UserS3Client.GetBaseClient()
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "ionoscloud_user_object_storage_bucket" {
+			continue
+		}
+		if rs.Primary.Attributes["name"] == "" {
```
adeatcu-ionos: `nitpick:` I don't think this check is necessary, the `name` is a `required`, `non-empty`, attribute.

internal/framework/services/userobjectstorage/resource_bucket_test.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("user object storage bucket still exists: %s", rs.Primary.Attributes["name"])
+	}
+
+	return nil
+}
+
+func TestAccUserObjectStorageBucketDataSource(t *testing.T) {
```
adeatcu-ionos: `nitpick:` I suggest including the data source test under the resource test in order to re-use the bucket instead of creating another one, although it doesn't matter because the provisioning time it's small.

internal/framework/services/userobjectstorage/resource_bucket_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
+		PreCheck: func() {
+			acctest.PreCheck(t)
+		},
+		CheckDestroy: testAccCheckUserBucketDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccUserBucketConfig_basic(rName),
```
adeatcu-ionos, line 32: I suggest adding a test in which we modify the value for the `force_destroy` attribute (setting it to `true`) and then check if the deletion was made properly.

services/bundleclient/bundleclient.go
```diff
@@ -55,6 +56,7 @@ func New(ctx context.Context, clientOptions clientoptions.TerraformClientOptions
 		VPNClient:         vpn.NewClient(clientOptions, fileConfig),
 		InMemoryDBClient:  inmemorydb.NewClient(clientOptions, fileConfig),
 		S3Client:          objectStorageService.NewClient(ctx, clientOptions, fileConfig),
+		UserS3Client:      userObjectStorageService.NewClient(ctx, clientOptions),
```
adeatcu-ionos, line 59: With the current code, we are initializing the client once and then we modify the URL based on the provided `region`. 

Because of this, if multiple buckets will be created in the same time, there will be synchronization issues that will lead to unexpected behavior.

I suggest the approach that we used lately: instantiate the client for each operation (create, update, etc.) based on the provided region.

adeatcu-ionos (changes requested):

internal/framework/services/userobjectstorage/data_source_bucket.go
cristiGuranIonos: it's used in framework for all resources from what I saw.

internal/framework/services/userobjectstorage/resource_bucket_test.go
cristiGuranIonos: will add later as the sdk is broken for now for uploading objects

internal/framework/services/userobjectstorage/resource_bucket.go
cristiGuranIonos: object lock is fixed, for region there is a problem with the API also.

services/userobjectstorage/client.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// ValidRegions holds the accepted region values for schema validation.
+var ValidRegions = []string{"de", "eu-central-2", "eu-south-2"}
+
+// DefaultRegion is used when no region is specified (e.g. during import).
+const DefaultRegion = "de"
+
+// NewClient creates a new user object storage client factory.
+func NewClient(ctx context.Context, clientOptions clientoptions.TerraformClientOptions) *Client {
```
adeatcu-ionos, line 47: This code structure is different from what we've done for other clients:
- get rid of client initialization inside `SdkBundle`;
- create a method for `SdkBundle` that uses `SdkBundle.clientOptions` and `SdkBundle.fileConfig` to initialize a specific client;

The good thing about this code structure is that we are re-using `signer` and `userAgent`, the bad thing is that we are not consistent.

I only wanted to point this out, please take this into consideration and then choose whatever option you consider it's the best.

services/userobjectstorage/client.go
```diff
@@ -0,0 +1,108 @@
...
+	))
+	userAgent := fmt.Sprintf(
+		"terraform-provider/%s_ionos-cloud-sdk-go-user-object-storage/%s_hashicorp-terraform/%s_terraform-plugin-sdk/%s_os/%s_arch/%s",
+		clientOptions.Version, userobjectstorage.Version, clientOptions.TerraformVersion,
+		meta.SDKVersionString(), runtime.GOOS, runtime.GOARCH, //nolint:staticcheck
+	)
+
+	return &Client{
+		clientOptions: clientOptions,
+		httpClient:    &http.Client{Transport: shared.CreateTransport(clientOptions.SkipTLSVerify, "")},
```
adeatcu-ionos: Shouldn't we use `clientOptions.Certificate` here instead of the empty string?

services/userobjectstorage/client.go
adeatcu-ionos, line 47: Do we want to add `file config` functionality?

adeatcu-ionos (changes requested):

docs/data-sources/user_object_storage_bucket.md
```diff
@@ -0,0 +1,38 @@
...
+  region = "de"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `name` - (Required)[string] The name of the bucket.
+- `region` - (Optional)[string] The region where the bucket resides. Defaults to `de` (Frankfurt). Valid values: `de`, `eu-central-2`, `eu-south-2`. Must match the bucket's actual region or the lookup will return not found.
```
adeatcu-ionos: I think it's a good idea to have a `read` timeout for the data source. I noticed that we are missing this functionality for some new resources, I will also create a JIRA task for this.

docs/resources/s3_bucket_cors_configuration.md
```diff
@@ -11,7 +11,7 @@ description: |-
 
 Manages Object Lock Configuration for Buckets on IonosCloud.
 
-⚠️ **Note:** The Terraform provider **only supports contract-owned buckets. User-owned buckets are not supported,** and there are no plans to introduce support for them. As a result, **user-owned buckets cannot be created, updated, deleted, read, or imported** using this provider.
+
```
adeatcu-ionos, line 14: `nitpick:` remove the new extra newline.

docs/resources/s3_bucket_lifecycle_configuration.md
```diff
@@ -11,7 +11,7 @@ description: |-
 
 Manages Lifecycle Configuration for Buckets on IonosCloud.
 
-⚠️ **Note:** The Terraform provider **only supports contract-owned buckets. User-owned buckets are not supported,** and there are no plans to introduce support for them. As a result, **user-owned buckets cannot be created, updated, deleted, read, or imported** using this provider.
+
```
adeatcu-ionos, line 14: `nitpick:` remove the new extra newline.

docs/resources/s3_bucket_object_lock_configuration.md
```diff
@@ -11,7 +11,7 @@ description: |-
 
 Manages Object Lock Configuration for Buckets on IonosCloud.
 
-⚠️ **Note:** The Terraform provider **only supports contract-owned buckets. User-owned buckets are not supported,** and there are no plans to introduce support for them. As a result, **user-owned buckets cannot be created, updated, deleted, read, or imported** using this provider.
+
```
adeatcu-ionos, line 14: `nitpick:` remove the new extra newline.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,252 @@
...
+		name = parts[1]
+	} else {
+		name = parts[0]
+	}
+	if name == "" {
+		resp.Diagnostics.AddError("invalid bucket name", "bucket name must not be empty")
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("region"), region)...)
```
adeatcu-ionos, line 290: If `region` is not used when importing, I think we have to set it to the default value here.

internal/framework/services/userobjectstorage/resource_bucket.go
```diff
@@ -0,0 +1,252 @@
...
+	readTimeout, diags := data.Timeouts.Read(ctx, utils.BucketDefaultTimeout)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, readTimeout)
+	defer cancel()
+
+	// Region cannot be read back from the API. Default when missing (plain `terraform import`).
+	if data.Region.ValueString() == "" {
```
adeatcu-ionos: I think this should be moved into the `create` method.

If the user specifies the `region` when creating the bucket => the state will have a value for `region`; in `read` we just read the state and set it back again so we are ok

If the user doesn't specify the `region` when creating the bucket, then, after bucket creation, set the `region` in the state to the default value => the state will have a value for `region`, same case as above.

services/userobjectstorage/bucket.go
```diff
@@ -0,0 +1,200 @@
...
+			req = req.ContinuationToken(continuationToken)
+		}
+
+		output, _, err := req.Execute()
+		if err != nil {
+			return fmt.Errorf("failed to list objects: %w", err)
+		}
+
+		for _, obj := range output.Contents {
+			if obj.Key == nil {
```
adeatcu-ionos, line 202: Isn't this an error case? 

If we retrieve a list of objects and, for some reason, one of the objects doesn't have a key it means that we can't delete that object so `deleteAllObjects` is failing.

services/userobjectstorage/bucket.go
```diff
@@ -0,0 +1,200 @@
...
+		return err
+	}
+
+	apiResp, err := api.BucketsApi.DeleteBucket(ctx, data.Name.ValueString()).Execute()
+	if apiResp.HttpNotFound() {
+		return nil
+	}
+
+	if isBucketNotEmptyError(err) && data.ForceDestroy.ValueBool() {
+		if err = c.deleteAllObjects(ctx, api, data.Name.ValueString()); err != nil {
```
adeatcu-ionos: The `delete` timeout doesn't include the time for the object deletion. It's worth noting this but I'm not suggesting any change because the timeout value is big enough.

services/userobjectstorage/bucket.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	if err != nil {
+		return backoff.Permanent(fmt.Errorf("failed to check if bucket was deleted: %w", err))
+	}
+	return fmt.Errorf("bucket still exists")
+}
+
+func isBucketNotEmptyError(err error) bool {
+	var apiErr shared.GenericOpenAPIError
+	return errors.As(err, &apiErr) && apiErr.StatusCode() == 409
```
adeatcu-ionos: We are matching the status code 409 but this status code represents several errors. 

Since this is the response for a `DELETE` request, I don't think there is a problem but you may want to consider this and maybe match the error message instead of the status code.

adeatcu-ionos (changes requested):

services/userobjectstorage/client.go
cristiGuranIonos: no, don't think will be needed

docs/data-sources/user_object_storage_bucket.md
cristiGuranIonos: added for read

internal/framework/services/userobjectstorage/resource_bucket.go
cristiGuranIonos: will make it required actually

internal/framework/services/userobjectstorage/resource_bucket.go
cristiGuranIonos: same, will make it required

services/userobjectstorage/client.go
cristiGuranIonos: will keep it as is

adeatcu-ionos (approved):

PR #990, opened 2026-05-20
feat: add query bucket and object storage access key

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


docs/list-resources/object_storage_accesskey.md
```diff
@@ -0,0 +1,52 @@
...
+subcategory: "Object storage management"
+layout: "ionoscloud"
+page_title: "IONOS CLOUD: object_storage_accesskey"
+description: |-
+  Lists IONOS CLOUD Object Storage Access Keys.
+---
+
+# List Resource: ionoscloud_object_storage_accesskey
+
+-> **Note:** List Resources require HashiCorp Terraform version 1.14 or later and are queried using `terraform query`.
```
adeatcu-ionos: `nitpick:` we usually use a warning sign when we are writing notes, not an arrow, I suggest doing the same here for consistency.

docs/list-resources/s3_bucket.md
```diff
@@ -0,0 +1,50 @@
...
+subcategory: "Object Storage"
+layout: "ionoscloud"
+page_title: "IONOS CLOUD: s3_bucket"
+description: |-
+  Lists IONOS Object Storage Buckets.
+---
+
+# List Resource: ionoscloud_s3_bucket
+
+-> **Note:** List Resources require HashiCorp Terraform version 1.14 or later and are queried using `terraform query`.
```
adeatcu-ionos: `nitpick:` warning sign instead of arrow for notes

docs/list-resources/object_storage_accesskey.md
```diff
@@ -0,0 +1,52 @@
...
+
+Lists [Object Storage Access Keys](https://docs.ionos.com/cloud/storage-and-backup/ionos-object-storage/concepts/key-management) on IONOS CLOUD.
+
+## Example Usage
+
+### List access keys
+
+```hcl
```
adeatcu-ionos, line 21: I'm not sure about this one but maybe it would be nice to specify that the list block should be placed in a file with `.tfquery.hcl` as an extension.

It's the customer's responsibility to know how to use the `query` functionality but since all the blocks from this documentation (until now) are written in a `.tf` file, maybe it makes sense to add this extra info to make things easier for the customer.

docs/resources/s3_bucket.md
```diff
@@ -57,6 +57,29 @@ The following arguments are supported:
 
 ## Import
 
+In Terraform v1.12.0 and later, the [`import` block](https://developer.hashicorp.com/terraform/language/import) can be used with the `identity` attribute. For example:
+
+```hcl
+import {
+  to = ionoscloud_s3_bucket.example
+  identity = {
```
adeatcu-ionos, line 65: I would also add an example here in which `region` is used.

internal/framework/identity/list.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	stream.Results = func(push func(list.ListResult) bool) {
+		for _, item := range items {
+			result := req.NewListResult(ctx)
+			mapped, diags := mapper(ctx, req.IncludeResource, item)
+			if diags.HasError() {
+				result.Diagnostics.Append(diags...)
+				push(result)
+				return
+			}
```
adeatcu-ionos: If `mapper` returns diagnostics that have warnings but no errors, they're never appended to `result` and are lost.


```suggestion
			mapped, diags := mapper(ctx, req.IncludeResource, item)
			result.Diagnostics.Append(diags...)
			if diags.HasError() {
				push(result)
				return
			}
```

internal/framework/identity/list.go
```diff
@@ -0,0 +1,49 @@
...
+		return
+	}
+	stream.Results = func(push func(list.ListResult) bool) {
+		for _, item := range items {
+			result := req.NewListResult(ctx)
+			mapped, diags := mapper(ctx, req.IncludeResource, item)
+			if diags.HasError() {
+				result.Diagnostics.Append(diags...)
+				push(result)
+				return
```
adeatcu-ionos: Below, in the same function, we are not stopping the execution if any error occurs during `result.Identity.Set` or `result.Resource.Set`, I think we should do the same here.

internal/framework/provider/provider.go
```diff
@@ -45,6 +46,11 @@ type FrameworkClientOptions struct {
 	Insecure       types.Bool   `tfsdk:"insecure"`
 }
 
+var (
+	_ provider.Provider                  = (*IonosCloudProvider)(nil)
+	_ provider.ProviderWithListResources = (*IonosCloudProvider)(nil)
```
adeatcu-ionos: `ProviderWithListResources` looks like this:
```
type ProviderWithListResources interface {
	Provider

	// ListResources returns a slice of functions to instantiate each Resource
	// List implementation.
	//
	// The resource type name is determined by the ListResource implementing
	// the Metadata method. All ListResources must have unique names.
	ListResources(context.Context) []func() list.ListResource
}
```

Therefore, the first line is redundant.

```suggestion
	_ provider.ProviderWithListResources = (*IonosCloudProvider)(nil)
```

internal/framework/services/objectstorage/resource_bucket_list.go
```diff
@@ -0,0 +1,89 @@
...
+
+// List lists all bucket resources.
+func (r *bucketResource) List(ctx context.Context, req list.ListRequest, stream *list.ListResultsStream) {
+	identity.StreamList(ctx, stream, req, r.client.ListBuckets, r.Map)
+}
+
+// Map returns a MappedItem describing the bucket, or nil to skip it.
+func (r *bucketResource) Map(ctx context.Context, includeResource bool, b objstorage.Bucket) (*identity.MappedItem, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	if b.Name == nil {
```
adeatcu-ionos, line 44: I think that  `b.Name == nil` should be an error case.

internal/framework/services/objectstorage/resource_bucket_list.go
```diff
@@ -0,0 +1,89 @@
...
+		Identity:    &identity.Model{ID: types.StringValue(*b.Name)},
+	}
+
+	if !includeResource {
+		return mapped, diags
+	}
+
+	region, err := r.client.GetBucketLocation(ctx, types.StringValue(*b.Name))
+	if err != nil {
+		diags.AddError("failed to get bucket location", err.Error())
```
adeatcu-ionos: Maybe include the bucket name in these error messages.

internal/framework/services/objectstorage/resource_bucket_test.go
```diff
@@ -37,6 +38,16 @@ func TestAccBucketResource(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "object_lock_enabled", "false"),
 				),
 			},
+			{
+				Query:  true,
+				Config: "list \"ionoscloud_s3_bucket\" \"test\" {\n  provider = ionoscloud\n}",
```
adeatcu-ionos, line 45: Is there any way to test the functionality for `include_resource = true`?

adeatcu-ionos (changes requested):

docs/list-resources/object_storage_accesskey.md
cristiGuranIonos: we actually use warning, > , sometimes nothing and once ~>. will change

internal/framework/services/objectstorage/resource_bucket_list.go
cristiGuranIonos: will remove it. it will never happen

internal/framework/services/objectstorage/resource_bucket_test.go
cristiGuranIonos: pipeline no, bucket added

internal/framework/identity/list.go
```diff
@@ -0,0 +1,56 @@
...
+		stream.Results = list.ListResultsStreamDiagnostics(diags)
+		return
+	}
+	stream.Results = func(push func(list.ListResult) bool) {
+		for _, item := range items {
+			result := req.NewListResult(ctx)
+			mapped, diags := mapper(ctx, req.IncludeResource, item)
+			result.Diagnostics.Append(diags...)
+			if mapped == nil {
+				continue
```
adeatcu-ionos, line 45: For buckets, if `mapped == nil` it means we are speaking about an error case, so the result will contain some diagnostics.

However, since we `continue` here without pushing the `result`, the diagnostics will be lost.

I suggest pushing the `result` here before continuing.

internal/framework/identity/list.go
```diff
@@ -0,0 +1,56 @@
...
+			}
+			if mapped.Identity == nil {
+				result.Diagnostics.AddError("mapper contract violation", "MappedItem.Identity must not be nil")
+				push(result)
+				return
```
adeatcu-ionos, line 50: As I said in the initial comment, I don't think we should stop the execution if any error occurs, I suggest using `continue` here instead of this `return` statement.

internal/framework/services/objectstorage/resource_bucket_list.go
```diff
@@ -0,0 +1,87 @@
...
+// List lists all bucket resources.
+func (r *bucketResource) List(ctx context.Context, req list.ListRequest, stream *list.ListResultsStream) {
+	identity.StreamList(ctx, stream, req, r.client.ListBuckets, r.Map)
+}
+
+// Map returns a MappedItem describing the bucket, or nil to skip it.
+func (r *bucketResource) Map(ctx context.Context, includeResource bool, b objstorage.Bucket) (*identity.MappedItem, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	region, err := r.client.GetBucketLocation(ctx, types.StringValue(*b.Name))
```
adeatcu-ionos, line 49: Dereferencing `b.Name` without checking for `nil` seems odd to me, I understand that it shouldn't happen but still.

adeatcu-ionos (changes requested):

internal/framework/identity/list.go
cristiGuranIonos: this is a hard error and always shows a programming error

internal/framework/identity/list.go
cristiGuranIonos: pushing here would be an invalid entry. It's not an error though, as this is a way to skip certain items.

internal/framework/services/objectstorage/resource_bucket_list.go
cristiGuranIonos: If this is not caught upstream, the S3 team has huge problems but I'll add it.

adeatcu-ionos (approved):

PR #993, opened 2026-05-28
Fea: add filtering for list for s3 bucke and acess keys

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Upcoming
 ### Features
 - New list resource: `ionoscloud_s3_bucket` — list all Object Storage buckets via `terraform query` (requires Terraform 1.14+).
 - New list resource: `ionoscloud_object_storage_accesskey` — list all Object Storage Access Keys via `terraform query` (requires Terraform 1.14+).
+- Add filters for `list` for s3 buckets and object storage access keys.
 
 ### Refactor
 - Refactor `ionoscloud_pg_cluster_v2` resource so `password` is now write-only, and require `password_version` to trigger password updates.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Upcoming
 ### Features
 - New list resource: `ionoscloud_s3_bucket` — list all Object Storage buckets via `terraform query` (requires Terraform 1.14+).
 - New list resource: `ionoscloud_object_storage_accesskey` — list all Object Storage Access Keys via `terraform query` (requires Terraform 1.14+).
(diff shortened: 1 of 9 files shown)


internal/framework/services/objectstorage/resource_bucket_test.go
```diff
@@ -68,6 +73,42 @@ func TestAccBucketResource(t *testing.T) {
 					),
 				},
 			},
+			{
+				Query: true,
```
adeatcu-ionos, line 80: In the current test configuration we have only one bucket. I suggest adding multiple buckets in order to be sure that the only bucket that is fetched is the one that respects the provided filters.

internal/framework/services/objectstorage/resource_bucket_list.go
```diff
@@ -29,7 +29,9 @@ func NewBucketListResource() list.ListResource {
 // ListResourceConfigSchema returns the schema for the configuration of the bucket list resource.
 func (r *bucketResource) ListResourceConfigSchema(_ context.Context, _ list.ListResourceSchemaRequest, resp *list.ListResourceSchemaResponse) {
 	resp.Schema = listschema.Schema{
-		Attributes: map[string]listschema.Attribute{},
+		Attributes: map[string]listschema.Attribute{
+			identity.FiltersKey: identity.FilterAttribute("name", "region"),
```
adeatcu-ionos, line 33: Why shouldn't we have multiple filters here? I'm thinking about `object_lock_enabled`, maybe even `tags`.

CHANGELOG.md
adeatcu-ionos:
```suggestion
## Upcoming
```

adeatcu-ionos:
```suggestion
## 6.7.29
```

adeatcu-ionos (changes requested):

internal/framework/services/objectstorage/resource_bucket_list.go
cristiGuranIonos: object_lock_enabled doesn't seem that useful to filter. Tags, not implemented right now as a first implementation. Can be added if anyone requests it.
We might look in the future at implementing regex search also for example if we get the request.
For now region for all buckets in a region, name for a unique bucket to import

internal/framework/services/objectstorage/resource_bucket_test.go
cristiGuranIonos: added

adeatcu-ionos (approved):

PR #995, opened 2026-06-05
feat: list for psqlv2 clusters

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated

### DIFF
--- a/docs/list-resources/psql_cluster.md
+++ b/docs/list-resources/psql_cluster.md
@@ -1,0 +1,127 @@
+---
+subcategory: "Database as a Service - PostgreSQL"
+layout: "ionoscloud"
+page_title: "IONOS CLOUD: pg_cluster_v2"
+description: |-
+  Lists IONOS CLOUD PostgreSQL v2 Clusters.
+---
+
+# List Resource: ionoscloud_pg_cluster_v2
+
+⚠️ **Note:** List Resources require HashiCorp Terraform version 1.14 or later and are queried using `terraform query`.
+
+Lists [IONOS PostgreSQL v2 Clusters](https://docs.ionos.com/cloud/databases/postgresql) on IONOS CLOUD.
+
+## Example Usage
+
+⚠️ **Note:** `list` blocks must be placed in a dedicated `tfquery.hcl` file, separate from your main Terraform configuration.
+
+### List all clusters
+
+```hcl
+list "ionoscloud_pg_cluster_v2" "all" {
+  provider         = ionoscloud
+  include_resource = true
+}
+```
+
+### Filter clusters by location
+
+```hcl
+list "ionoscloud_pg_cluster_v2" "de_txl" {
+  provider         = ionoscloud
+  include_resource = true
+  config {
+    filters = [{
+      field_name  = "location"
+      field_value = "de/txl"
+    }]
+  }
+}
+```
+
+### Filter clusters by name and location
+
+```hcl
+list "ionoscloud_pg_cluster_v2" "prod" {
+  provider         = ionoscloud
+  include_resource = true
+  config {
+    filters = [
+      { field_name = "name",     field_value = "my-cluster" },
+      { field_name = "location", field_value = "de/txl" },
+    ]
+  }
+}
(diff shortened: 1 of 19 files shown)


rmocanu-ionos (dismissed):

docs/list-resources/psql_cluster.md
```diff
@@ -0,0 +1,127 @@
+---
+subcategory: "Database as a Service - PostgreSQL"
```
adeatcu-ionos: @cristiGuranIonos shouldn't this file be named `psql_cluster_v2` ?

docs/list-resources/psql_cluster_v2.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+Terraform will write an `ionoscloud_pg_cluster_v2` resource block for each discovered cluster into `imported.tf`, which can then be used directly in your configuration.
+
+## Argument Reference
+
+The `config` block supports the following arguments:
+
+- `filters` - (Optional) List of filters to apply. All filters must match (AND logic). Each filter supports:
+  - `field_name` - (Required) The field to filter on. Supported values: `name`, `location`.
```
adeatcu-ionos, line 73: @cristiGuranIonos why do we mark these as `Required` in the documentation? I notice we have the same for S3 buckets.

Below we have:

```
**Performance note:** When no `location` filter is set,
```

This means that the location filter is not required.

docs/resources/pg_cluster_v2.md
```diff
@@ -137,3 +137,31 @@ Resource DBaaS PostgreSQL v2 Cluster can be imported using the format `location:
...
+  config {
+    filters = [
+      { field_name = "name",     field_value = "my-cluster" },
+      { field_name = "location", field_value = "de/txl" },
+    ]
+  }
+}
+```
+
+See the [ionoscloud_pg_cluster_v2 list resource](../list-resources/psql_cluster.md) documentation for the full filter reference.
```
adeatcu-ionos: @cristiGuranIonos since we added identity, shouldn't we update this doc with the new way to import the cluster?

services/bundleclient/bundleclient.go
```diff
@@ -161,19 +161,19 @@ func (c SdkBundle) NewContainerRegistryClient(ctx context.Context, location stri
 	}
 	config.HTTPClient = &http.Client{}
 	config.HTTPClient.Transport = shared.CreateTransport(endpoint.SkipTLSVerify, endpoint.CertificateAuthData)
-	return crService.NewClientFromConfig(config), nil
+	return crservice.NewClientFromConfig(config), nil
 }
 
 // NewMongoClient creates a new MongoDB client for a specific location
-func (c SdkBundle) NewMongoClient(ctx context.Context, location string) (*dbaasService.MongoClient, error) {
+func (c SdkBundle) NewMongoClient(ctx context.Context, location string) (*dbaasservice.MongoClient, error) {
```
adeatcu-ionos, line 168: @cristiGuranIonos I would do these in another PR so we don't mix things up. We still have to rename a lot of packages I think.

rmocanu-ionos (dismissed):

docs/list-resources/psql_cluster_v2.md
cristiGuranIonos: The filter is optional, but if given the field_name and field_value are required.

adeatcu-ionos: Don't mind my comment, I made a confusion, sorry!

docs/resources/pg_cluster_v2.md
cristiGuranIonos: updated other resources with list with additional details on how to list

services/bundleclient/bundleclient.go
cristiGuranIonos: This is related to dbaas so we do it now, we can do the rest in another PR.

adeatcu-ionos (approved):

PR #1002, opened 2026-06-15
refactor: get image function

cristiGuranIonos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Jira task updated


.github/workflows/test.yml
```diff
@@ -48,10 +48,17 @@ jobs:
...
         if: ${{ github.event.inputs.test-tags != '' && github.events.inputs.failfast == false }}
-        run: go test ./... -v -timeout 6h -tags ${{ github.event.inputs.test-tags }}
+        run: go test ./... -v -timeout 6h -tags ${{ github.event.inputs.test-tags }} 2>&1 | tee /tmp/test_output.txt; exit ${PIPESTATUS[0]}
       - name: Run tests without tags
         if: ${{ github.event.inputs.test-tags == ''}}
-        run: go test ./... -v -failfast -timeout 180m
\ No newline at end of file
+        run: go test ./... -v -failfast -timeout 180m 2>&1 | tee /tmp/test_output.txt; exit ${PIPESTATUS[0]}
+      - name: Show failed tests
+        if: always()
```
adeatcu-ionos: @cristiGuranIonos please consider using `failure() ` here. I understand that `always()` will run the step even when we cancel the run and this may be confusing.

ionoscloud/resource_volume.go
```diff
@@ -760,172 +760,148 @@ func getVolumeData(ctx context.Context, d *schema.ResourceData, path, serverType
...
+			if hasImageCredentials(volume) {
+				return "", "", fmt.Errorf("passwords/SSH keys are not supported for snapshots")
+			}
+			return image, "", nil
+		}
+		imageAlias = cloudapiimage.GetImageAlias(imageName, images, locationIDs)
+		if imageAlias == "" {
+			if rejectedImage != nil {
+				return "", "", fmt.Errorf(
+					"image '%s' was found (name: '%s') with type '%s' in location '%s'; "+
```
adeatcu-ionos, line 830: @cristiGuranIonos I don't think it makes sense to print the name twice (`imageName` will never be different from `*rejectedImage.Properties.Name`, exact match is done).

ionoscloud/resource_volume.go
```diff
@@ -760,172 +760,148 @@ func getVolumeData(ctx context.Context, d *schema.ResourceData, path, serverType
...
+func getImageByUUID(ctx context.Context, client *ionoscloud.APIClient, dcID, imageID string, volume ionoscloud.VolumeProperties) (resolvedImage string, err error) {
+	img, apiResponse, err := client.ImagesApi.ImagesFindById(ctx, imageID).Execute()
+	logApiRequestTime(apiResponse)
+	if apiResponse.HttpNotFound() {
+		snapshot, apiResponse, err := client.SnapshotsApi.SnapshotsFindById(ctx, imageID).Execute()
+		apiResponse.LogInfo()
+		if err != nil {
+			return "", fmt.Errorf("could not fetch image/snapshot: %w", err)
+		}
+		if hasImageCredentials(volume) {
```
adeatcu-ionos, line 855: @cristiGuranIonos I think we should check this before `SnapshotsFindById` call.

adeatcu-ionos (changes requested):

ionoscloud/resource_volume.go
cristiGuranIonos: I think these can differ actually. As I remember I saw the err message in prod and they were different.

ionoscloud/resource_volume.go
cristiGuranIonos: It doesn't matter, as it's a ById it won't take a lot of time. It would also be a bit more accurate, if snapshot exists, then we validate that it has credentials.

.github/workflows/test.yml
cristiGuranIonos: good idea

adeatcu-ionos (approved):

PR #1029, opened 2026-08-19
feat: add support for the de/fra/1 location

mimihalescu:
Adds `de/fra/1` (Frankfurt West) to every regional endpoint map and available-locations list. Like `de/fra/2` it is a facility inside the `de/fra` metro region and shares the same regional endpoint.

The PgSQL location drift suppression is generalised behind a new `IsChildOf` helper so it covers any child location rather than only `de/fra/2`.

Two long-standing issues in the same lists are fixed along the way: `ValidNFSLocations` spelled `fr/par` and `gb/lhr` with dashes so they never matched the endpoint map, and the unreferenced `constant.Locations` slice is removed.


docs/data-sources/logging_pipeline.md
```diff
@@ -33,7 +33,7 @@ data "ionoscloud_logging_pipeline" "example" {
 ```
 
 ## Argument reference
-* `location` - (Optional)[string] The location of the Logging pipeline. Default: `de/txl`, other available locations: `de/fra`, `de/fra/2`, `de/txl`, `es/vit`, `gb/bhx`, `gb/lhr`,  `fr/par`, `us/mci`.
+* `location` - (Optional)[string] The location of the Logging pipeline. Default: `de/txl`, other available locations: `de/fra`, `de/fra/1`, `de/fra/2`, `de/txl`, `es/vit`, `gb/bhx`, `gb/lhr`,  `fr/par`, `us/mci`.
```
adeatcu-ionos: @mimihalescu as Copilot says there are some inconsistencies between the locations mentioned in the doc file and the locations supported in the client. Other than these, the PR looks good.

adeatcu-ionos (changes requested):

docs/data-sources/logging_pipeline.md
mimihalescu: fixed in a1e6db2

adeatcu-ionos (approved):

PR #1035, opened 2026-08-31
fix: enabled_features for vcpu_server

### LINKED ISSUE #1033: bug: ionoscloud_vcpu_server fails with "error setting enabled_features Invalid address to set" on provider 6.7.36
## Description
  Upgrading the provider from 6.7.35 to 6.7.36 breaks `ionoscloud_vcpu_server` resources during plan/apply/refresh. Every operation fails with:

  Error: error setting enabled_features Invalid address to set: []string{"enabled_features"}

  Root cause: PR #1028 ("feat: confidential compute") added an `enabled_features` field to the shared server state-writer (`setResourceServerData` in `resource_server.go`), which is also used by `resourceVCPUServerCreate`/`resourceServerRead` for the `ionoscloud_vcpu_server` resource.
The PR added the `enabled_features` schema attribute to `ionoscloud_server` and to the  `ionoscloud_vcpu_server` **data source**, but not to the `ionoscloud_vcpu_server` **resource** schema (`resource_vcpu_server.go`). The shared reader unconditionally calls  `d.Set("enabled_features", ...)`, which fails because that key doesn't exist in the resource's schema.

  This affects every `ionoscloud_vcpu_server` resource, regardless of confidential-compute usage.

  ## Expected behavior
  `ionoscloud_vcpu_server` resources plan/apply/refresh successfully on 6.7.36, same as on 6.7.35.

  ## Configuration Files

  ```hcl
  resource "ionoscloud_vcpu_server" "name" {
{...}
  }
 ```

 ## How to Reproduce

  1. Define any ionoscloud_vcpu_server resource.
  2. Set provider version to 6.7.36.
  3. Run terraform plan (or apply/refresh).

  ## Error and Debug Output
  │ Error: error setting enabled_features Invalid address to set: []string{"enabled_features"}
  │ Resource ID: 
  │ Resource Name:
  │ Contract number: 
  │
  │   with ionoscloud_vcpu_server.mainframe,
  This affects every `ionoscloud_vcpu_server` resource, regardless of confidential-compute
  usage.

 ## Additional Notes

  Workaround: pin provider to 6.7.35 until fixed.

  Fix should add an enabled_features (Computed, TypeList/TypeString as appropriate) attribute to
  the ionoscloud_vcpu_server resource schema in resource_vcpu_server.go, mirroring what was
  already done for the data source and for ionoscloud_server.

  ## References

  - Introduced by commit a3f6453700ab61957c8aa3ca53a84aef32034218 / PR #1028 ("feat:
  confidential compute")

avirtopeanu-ionos:
fix #1033: `error setting enabled_features` on `vcpu_server` by adding the missing `enabled_features` (Computed) attribute to the resource schema, introduced in 6.7.36 by the Confidential Computing change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 6.7.37
 
 ### Fixes
-- `ionoscloud_vcpu_server`: Fix `error setting enabled_features` on every plan/apply/refresh by adding the missing `enabled_features` (Computed) attribute to the resource schema, introduced in 6.7.36 by the Confidential Computing change.
+- `ionoscloud_vcpu_server`: Fix `error setting enabled_features` on every plan/apply/refresh by adding the missing `enabled_features` and `confidential` (both Computed) attributes to the resource schema. The shared server reader (`setResourceServerData`) sets both unconditionally, so both keys must exist; the crash was introduced in 6.7.36 by the Confidential Computing change.
 
 ## 6.7.36
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 6.7.37
+
+### Fixes
+- `ionoscloud_vcpu_server`: Fix `error setting enabled_features` on every plan/apply/refresh by adding the missing `enabled_features` (Computed) attribute to the resource schema, introduced in 6.7.36 by the Confidential Computing change.
+
 ## 6.7.36
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 6.7.37
 
 ### Fixes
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 6.7.37
+
+### Fixes
+- `ionoscloud_vcpu_server`: Fix `error setting enabled_features` on every plan/apply/refresh by adding the missing `enabled_features` and `confidential` (both Computed) attributes to the resource schema. The shared server reader (`setResourceServerData`) sets both unconditionally, so both keys must exist; the crash was introduced in 6.7.36 by the Confidential Computing change.
```
cristiGuranIonos: would it make sense to no longer have a shared setResourceServerData?

avirtopeanu-ionos: sure, I can refactor this so it is not shared anymore

cristiGuranIonos (approved):

